PR #1, opened 2024-02-20
Upgrade to latest Smithay

ABeltramo:
This is mainly in order to add support for the [pointer constraints](https://wayland.app/protocols/pointer-constraints-unstable-v1) Wayland protocol so that SDL will eventually pick up the relative mouse movements when going outside the edge of the window.

 This turns out to be a huge refactor..

### DIFF
--- a/wayland-display-core/src/comp/focus.rs
+++ b/wayland-display-core/src/comp/focus.rs
@@ -1,3 +1,4 @@
+use gst_video::prelude::VideoDecoderExtManual;
 use smithay::{
     backend::input::KeyState,
     desktop::{PopupKind, Window},
@@ -10,6 +11,8 @@ use smithay::{
     utils::{IsAlive, Serial},
     wayland::seat::WaylandFocus,
 };
+use smithay::input::pointer::{GestureHoldBeginEvent, GestureHoldEndEvent, GesturePinchBeginEvent, GesturePinchEndEvent, GesturePinchUpdateEvent, GestureSwipeBeginEvent, GestureSwipeEndEvent, GestureSwipeUpdateEvent};
+use crate::comp::State;
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum FocusTarget {
@@ -128,6 +131,69 @@ impl PointerTarget<super::State> for FocusTarget {
         match self {
             FocusTarget::Wayland(w) => w.axis(seat, data, frame),
             FocusTarget::Popup(p) => p.wl_surface().axis(seat, data, frame),
+        }
+    }
+
+    fn frame(&self, seat: &Seat<State>, data: &mut State) {
+        match self {
+            FocusTarget::Wayland(w) => w.frame(seat, data),
+            FocusTarget::Popup(p) => {todo!()},
+        }
+    }
+
+    fn gesture_swipe_begin(&self, seat: &Seat<State>, data: &mut State, event: &GestureSwipeBeginEvent) {
+        match self {
+            FocusTarget::Wayland(w) => w.gesture_swipe_begin(seat, data, event),
+            FocusTarget::Popup(p) => p.wl_surface().gesture_swipe_begin(seat, data, event),
+        }
+    }
+
+    fn gesture_swipe_update(&self, seat: &Seat<State>, data: &mut State, event: &GestureSwipeUpdateEvent) {
+        match self {
+            FocusTarget::Wayland(w) => w.gesture_swipe_update(seat, data, event),
+            FocusTarget::Popup(p) => p.wl_surface().gesture_swipe_update(seat, data, event),
+        }
+    }
+
+    fn gesture_swipe_end(&self, seat: &Seat<State>, data: &mut State, event: &GestureSwipeEndEvent) {
+        match self {
+            FocusTarget::Wayland(w) => w.gesture_swipe_end(seat, data, event),
+            FocusTarget::Popup(p) => p.wl_surface().gesture_swipe_end(seat, data, event),
+        }
+    }
+
+    fn gesture_pinch_begin(&self, seat: &Seat<State>, data: &mut State, event: &GesturePinchBeginEvent) {
+        match self {
+            FocusTarget::Wayland(w) => w.gesture_pinch_begin(seat, data, event),
+            FocusTarget::Popup(p) => p.wl_surface().gesture_pinch_begin(seat, data, event),
+        }
+    }
+
+    fn gesture_pinch_update(&self, seat: &Seat<State>, data: &mut State, event: &GesturePinchUpdateEvent) {
+        match self {
+            FocusTarget::Wayland(w) => w.gesture_pinch_update(seat, data, event),
+            FocusTarget::Popup(p) => p.wl_surface().gesture_pinch_update(seat, data, event),
+        }
+    }
+
+    fn gesture_pinch_end(&self, seat: &Seat<State>, data: &mut State, event: &GesturePinchEndEvent) {
+        match self {
+            FocusTarget::Wayland(w) => w.gesture_pinch_end(seat, data, event),
+            FocusTarget::Popup(p) => p.wl_surface().gesture_pinch_end(seat, data, event),
+        }
+    }
+
+    fn gesture_hold_begin(&self, seat: &Seat<State>, data: &mut State, event: &GestureHoldBeginEvent) {
+        match self {
+            FocusTarget::Wayland(w) => w.gesture_hold_begin(seat, data, event),
+            FocusTarget::Popup(p) => p.wl_surface().gesture_hold_begin(seat, data, event),
+        }
+    }
+
+    fn gesture_hold_end(&self, seat: &Seat<State>, data: &mut State, event: &GestureHoldEndEvent) {
+        match self {
+            FocusTarget::Wayland(w) => w.gesture_hold_end(seat, data, event),
+            FocusTarget::Popup(p) => p.wl_surface().gesture_hold_end(seat, data, event),
         }
     }
 
--- a/wayland-display-core/src/comp/focus.rs
+++ b/wayland-display-core/src/comp/focus.rs
@@ -41,11 +41,11 @@ impl From<PopupKind> for FocusTarget {
     }
 }
 
-impl KeyboardTarget<super::State> for FocusTarget {
+impl KeyboardTarget<State> for FocusTarget {
     fn enter(
         &self,
-        seat: &Seat<super::State>,
-        data: &mut super::State,
+        seat: &Seat<State>,
+        data: &mut State,
         keys: Vec<KeysymHandle<'_>>,
         serial: Serial,
     ) {
@@ -57,7 +57,7 @@ impl KeyboardTarget<super::State> for FocusTarget {
         }
     }
 
-    fn leave(&self, seat: &Seat<super::State>, data: &mut super::State, serial: Serial) {
+    fn leave(&self, seat: &Seat<State>, data: &mut State, serial: Serial) {
         match self {
             FocusTarget::Wayland(w) => KeyboardTarget::leave(w, seat, data, serial),
             FocusTarget::Popup(p) => KeyboardTarget::leave(p.wl_surface(), seat, data, serial),
@@ -66,8 +66,8 @@ impl KeyboardTarget<super::State> for FocusTarget {
 
     fn key(
         &self,
-        seat: &Seat<super::State>,
-        data: &mut super::State,
+        seat: &Seat<State>,
+        data: &mut State,
         key: KeysymHandle<'_>,
         state: KeyState,
         serial: Serial,
@@ -81,8 +81,8 @@ impl KeyboardTarget<super::State> for FocusTarget {
 
     fn modifiers(
         &self,
-        seat: &Seat<super::State>,
-        data: &mut super::State,
+        seat: &Seat<State>,
+        data: &mut State,
         modifiers: ModifiersState,
         serial: Serial,
     ) {
@@ -93,15 +93,15 @@ impl KeyboardTarget<super::State> for FocusTarget {
     }
 }
 
-impl PointerTarget<super::State> for FocusTarget {
-    fn enter(&self, seat: &Seat<super::State>, data: &mut super::State, event: &MotionEvent) {
+impl PointerTarget<State> for FocusTarget {
+    fn enter(&self, seat: &Seat<State>, data: &mut State, event: &MotionEvent) {
         match self {
             FocusTarget::Wayland(w) => PointerTarget::enter(w, seat, data, event),
             FocusTarget::Popup(p) => PointerTarget::enter(p.wl_surface(), seat, data, event),
         }
     }
 
-    fn motion(&self, seat: &Seat<super::State>, data: &mut super::State, event: &MotionEvent) {
+    fn motion(&self, seat: &Seat<State>, data: &mut State, event: &MotionEvent) {
         match self {
             FocusTarget::Wayland(w) => w.motion(seat, data, event),
             FocusTarget::Popup(p) => p.wl_surface().motion(seat, data, event),
@@ -110,8 +110,8 @@ impl PointerTarget<super::State> for FocusTarget {
 
     fn relative_motion(
         &self,
-        seat: &Seat<super::State>,
-        data: &mut super::State,
+        seat: &Seat<State>,
+        data: &mut State,
         event: &RelativeMotionEvent,
     ) {
         match self {
@@ -120,14 +120,14 @@ impl PointerTarget<super::State> for FocusTarget {
         }
     }
 
-    fn button(&self, seat: &Seat<super::State>, data: &mut super::State, event: &ButtonEvent) {
+    fn button(&self, seat: &Seat<State>, data: &mut State, event: &ButtonEvent) {
         match self {
             FocusTarget::Wayland(w) => w.button(seat, data, event),
             FocusTarget::Popup(p) => p.wl_surface().button(seat, data, event),
         }
     }
 
-    fn axis(&self, seat: &Seat<super::State>, data: &mut super::State, frame: AxisFrame) {
+    fn axis(&self, seat: &Seat<State>, data: &mut State, frame: AxisFrame) {
         match self {
             FocusTarget::Wayland(w) => w.axis(seat, data, frame),
             FocusTarget::Popup(p) => p.wl_surface().axis(seat, data, frame),
@@ -137,7 +137,7 @@ impl PointerTarget<super::State> for FocusTarget {
     fn frame(&self, seat: &Seat<State>, data: &mut State) {
         match self {
             FocusTarget::Wayland(w) => w.frame(seat, data),
-            FocusTarget::Popup(p) => {todo!()},
+            FocusTarget::Popup(p) => { todo!("the trait `PointerTarget<_>` is not implemented for `&WlSurface`") }
         }
     }
 
@@ -197,7 +197,7 @@ impl PointerTarget<super::State> for FocusTarget {
         }
     }
 
-    fn leave(&self, seat: &Seat<super::State>, data: &mut super::State, serial: Serial, time: u32) {
+    fn leave(&self, seat: &Seat<State>, data: &mut State, serial: Serial, time: u32) {
         match self {
             FocusTarget::Wayland(w) => PointerTarget::leave(w, seat, data, serial, time),
             FocusTarget::Popup(p) => PointerTarget::leave(p.wl_surface(), seat, data, serial, time),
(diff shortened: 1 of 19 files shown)


wayland-display-core/src/wayland/handlers/compositor.rs
```diff
@@ -26,6 +28,10 @@ impl CompositorHandler for State {
         &mut self.compositor_state
     }
 
+    fn client_compositor_state<'a>(&self, client: &'a Client) -> &'a CompositorClientState {
+        todo!()
```
ABeltramo: This trait was required now, but I'm not sure what the implementation expects here. Should I return an empty `CompositorClientState`?

wayland-display-core/src/comp/focus.rs
```diff
@@ -131,6 +134,69 @@ impl PointerTarget<super::State> for FocusTarget {
         }
     }
 
+    fn frame(&self, seat: &Seat<State>, data: &mut State) {
+        match self {
+            FocusTarget::Wayland(w) => w.frame(seat, data),
+            FocusTarget::Popup(p) => {todo!()},
```
ABeltramo: `p.wl_surface().frame()` expects `frame_number` which I'm not sure how to extract in here..

wayland-display-core/src/comp/input.rs
```diff
@@ -23,14 +24,14 @@ use std::{
...
 
 pub struct NixInterface;
 
 impl LibinputInterface for NixInterface {
     fn open_restricted(&mut self, path: &Path, flags: i32) -> Result<OwnedFd, i32> {
-        fcntl::open(path, OFlag::from_bits_truncate(flags), stat::Mode::empty())
-            .map(|fd| unsafe { OwnedFd::from_raw_fd(fd) })
-            .map_err(|err| err as i32)
+        open(path, OFlags::from_bits_truncate(flags as u32), Mode::empty())
+            .map_err(|err| todo!("Convert Errno to i32?"))
```
ABeltramo: This is unfortunately just me being not much confident in Rust, I'm sure it's easily done..

wayland-display-core/src/comp/rendering.rs
```diff
@@ -1,24 +1,32 @@
...
+use smithay::backend::renderer::element::memory::MemoryBuffer;
+use smithay::reexports::drm::buffer::DrmFourcc;
+use smithay::utils::Size;
 
 use super::State;
 
-pub const CURSOR_DATA_BYTES: &[u8] = include_bytes!("../../resources/cursor.rgba");
+pub const CURSOR_DATA_BYTES: MemoryBuffer = MemoryBuffer::from_slice(
+    include_bytes!("../../resources/cursor.rgba"),
+    Fourcc::Rgba4444, // TODO: Format???
```
ABeltramo: Which format is the cursor? Btw, is this the right way to load it up in the new `MemoryBuffer` struct?

wayland-display-core/src/comp/rendering.rs
```diff
@@ -45,38 +53,43 @@ impl State {
...
-                )],
-                CursorImageStatus::Surface(wl_surface) => {
-                    smithay::backend::renderer::element::surface::render_elements_from_surface_tree(
-                        &mut self.renderer,
-                        wl_surface,
-                        self.pointer_location.to_physical_precise_round(1),
-                        1.,
-                    )
+                    CursorImageStatus::Named(cursor_icon) => vec![CursorElement::Memory(
+                        // TODO: icon?
```
ABeltramo, line 49: I saw that there's now this new `Named` instead of `Default` and there's a `cursor_icon` param that goes along with it; I'm not sure where to plug it..

wayland-display-core/src/comp/rendering.rs
```diff
@@ -91,7 +104,7 @@ impl State {
                     self.video_info.as_ref().unwrap().width() as i32,
                     self.video_info.as_ref().unwrap().height() as i32,
                 ),
-            ))
+            ), DrmFourcc::Rgba4444) // TODO: format???
```
ABeltramo: Which format should I pick here?

wayland-display-core/src/comp/rendering.rs
```diff
@@ -107,15 +120,15 @@ impl State {
...
                 let plane_data = vframe.plane_data_mut(0).unwrap();
                 plane_data.clone_from_slice(map);
             }
 
             buffer
         };
         self.renderer.unbind().map_err(DTRError::Rendering)?;
-
-        Ok((buffer, damage, render_element_states))
+        todo!("Get damage!");
```
ABeltramo: Previously `damage` was returned by the `render_output` method, how can I get that here?

wayland-display-core/src/wayland/handlers/relative_pointer.rs
```diff
@@ -1,5 +1,17 @@
...
+use smithay::wayland::pointer_constraints::PointerConstraintsHandler;
 
 use crate::comp::State;
 
-delegate_relative_pointer!(State);
\ No newline at end of file
+impl OutputHandler for State {}
+impl PointerConstraintsHandler for State{
+    fn new_constraint(&mut self, surface: &WlSurface, pointer: &PointerHandle<Self>) {
+        todo!()
```
ABeltramo: I guess this is what needs to be implemented in order to support the pointer constraints protocol; not sure what is expected in that method..

wayland-display-core/src/wayland/handlers/xdg.rs
```diff
@@ -78,6 +78,10 @@ impl XdgShellHandler for State {
             }
         }
     }
+
+    fn reposition_request(&mut self, surface: PopupSurface, positioner: PositionerState, token: u32) {
+        todo!()
```
ABeltramo, line 86: This traits was required now, I'm not sure how to implement it..

wayland-display-core/src/wayland/protocols/wl_drm.rs
```diff
@@ -156,29 +156,30 @@ where
...
-                                );
-                            }
-
-                            Err(ImportError::Failed) => {
-                                // Buffer import failed. The protocol documentation heavily implies killing the
-                                // client is the right thing to do here.
-                                drm.post_error(wl_drm::Error::InvalidName, "buffer import failed");
-                            }
-                        }
+                        todo!("How can I make a ImportNotifier? Constructor is private..");
```
ABeltramo: I've changed  `DmabufHandler::dmabuf_imported` so that it takes an `ImportNotifier` in input instead of returning the result like it used to; it seems that I can't find a way to create it though; the constructor is private, what am I missing?

ABeltramo (commented):
I've just done a first pass of things that were simply broken and added a few `todo` here and there to try and get along. The code is still not compiling, so this is far from over but I wanted to start getting some feedback early instead of deep diving further.
@Drakulix when you got time, please send some help 😅

wayland-display-core/src/comp/focus.rs
Drakulix: It is probably defaulting to the wrong `frame`-method, try:
```suggestion
            FocusTarget::Popup(p) => PointerTarget::frame(&p.wl_surface(), seat, data),
```

wayland-display-core/src/comp/input.rs
Drakulix:
```suggestion
            .map_err(|err| err.raw_os_error())
```

wayland-display-core/src/comp/input.rs
```diff
@@ -60,7 +61,7 @@ impl State {
                                 && !modifiers.alt
                                 && !modifiers.logo
                             {
-                                match handle.modified_sym() {
+                                match u32::from(handle.modified_sym()) {
```
Drakulix: Would be cleaner to not change this line and match on the new `Keysym` type instead: https://smithay.github.io/smithay/smithay/input/keyboard/struct.Keysym.html

wayland-display-core/src/comp/input.rs
```diff
@@ -60,7 +61,7 @@ impl State {
...
                                     keysyms::KEY_Tab => {
```
Drakulix: So this would become 
```suggestion
                                    Keysym::Tab => {
```

and so on...

wayland-display-core/src/comp/input.rs
```diff
@@ -94,7 +95,7 @@ impl State {
                                 }
                             }
                         } else {
-                            if data.surpressed_keys.remove(&handle.modified_sym()) {
+                            if data.surpressed_keys.remove(&(u32::from(handle.modified_sym()))) {
```
Drakulix: Little cleaner:
```suggestion
                            if data.surpressed_keys.remove(&handle.modified_sym().raw()) {
```

wayland-display-core/src/comp/input.rs
```diff
@@ -192,29 +193,29 @@ impl State {
                 self.last_pointer_movement = Instant::now();
                 let horizontal_amount = event
                     .amount(Axis::Horizontal)
-                    .or_else(|| event.amount_discrete(Axis::Horizontal).map(|x| x * 2.0))
+                    .or_else(|| event.amount_v120(Axis::Horizontal).map(|x| x * 2.0))
```
Drakulix:
```suggestion
                    .or_else(|| event.amount_v120(Axis::Horizontal).map(|x| x * 2.0 / 120.0))
```

The new v120 are different.

wayland-display-core/src/comp/input.rs
```diff
@@ -192,29 +193,29 @@ impl State {
...
                     .unwrap_or(0.0);
                 let vertical_amount = event
                     .amount(Axis::Vertical)
-                    .or_else(|| event.amount_discrete(Axis::Vertical).map(|y| y * 2.0))
+                    .or_else(|| event.amount_v120(Axis::Vertical).map(|y| y * 2.0))
```
Drakulix:
```suggestion
                    .or_else(|| event.amount_v120(Axis::Vertical).map(|y| y * 2.0 / 120.0))
```

wayland-display-core/src/comp/rendering.rs
Drakulix: Either `Rgba8888` or `Abgr8888`, I don't fully remember, probably the former given the name, but the different apis use different edianess to express the order...

wayland-display-core/src/comp/rendering.rs
Drakulix: Named means the application just provided a name of an icon to load from the cursor theme. We don't have that right now, so skipping that is *a solution*.

For a proper one check out: https://github.com/Smithay/smithay/blob/master/anvil/src/cursor.rs

wayland-display-core/src/comp/rendering.rs
Drakulix: Theoretically whatever is in `self.video_info`.

But OpenGL doesn't give us a bunch of formats and the src doesn't advertise anything but 32-bit Rgbx anyway. But the Fourcc-endianess is reversed and we need the alpha variant so `Abgr8888` is the correct one.

wayland-display-core/src/comp/rendering.rs
Drakulix: `render_output` returns a `RenderOutputResult` now, which contains both the `damage` and the `render_elements_states` now: https://smithay.github.io/smithay/smithay/backend/renderer/damage/struct.RenderOutputResult.html

It *also* contains a sync-point now, which we would need to wait on before giving a hardware buffer to gstreamer or we might not be done writing to it (or even better give that and the dmabuf to gstreamer, but that api change will take a another few years probably...).
But because we copy to memory anyway, OpenGL should already synchronize for us and we can just ignore it.

wayland-display-core/src/wayland/handlers/compositor.rs
Drakulix: No, we need to initialize that on client creation, store that in the Client and retrieve it (e.g. anvil: https://github.com/Smithay/smithay/blob/master/anvil/src/shell/mod.rs#L104-L106).

For initialization it should be enough to add a `CompositorClientState` to our `ClientState` struct, as that implements `Default`.

wayland-display-core/src/wayland/handlers/relative_pointer.rs
Drakulix: Store that in the state somewhere and then obey the constraint in the input code.

See the PR that implemented this in anvil: https://github.com/Smithay/smithay/pull/872/commits/d978a2b

Also we need to constuct the `PointerConstraintsState` in `src/comp/mod.rs:init` or we never advertise support for it.

(And it would be nice, if this would be put into a separate file like all the other protocol handlers.)

wayland-display-core/src/wayland/handlers/xdg.rs
Drakulix: We don't properly compute pointer constraints in this compositor at all, so we can't really do this.

The good thing is, this used to be much more complicated, but a lot of helpers were recently implemented properly in smithay. Just copy what smallvil does: https://github.com/Smithay/smithay/pull/1256/files#diff-f471a178782c2c9f3ca35beb0d6d87a219705ba01eaa0f9f1489916f44a7fbcc

Which just makes sure the popup is always visible on the current output (which is exactly what we need for our usecase).

wayland-display-core/src/wayland/protocols/wl_drm.rs
Drakulix: Yeah, sadly this is the case, so we can't piggy back on the existing `DmabufHandler` implementation any more.

So we either need our own trait and implement that as well (e.g. see how cosmic-comp does this now: https://github.com/pop-os/cosmic-comp/blob/master/src/wayland/protocols/drm.rs#L47) or just drop the `D` and the split between `protocols` and `handler` and just hard-code everything for our state struct, so that we can just call `state.renderer.import_dmabuf` directly here.

Drakulix (commented):
Looks pretty good already!

wayland-display-core/src/wayland/handlers/wl_drm.rs
```diff
@@ -1,3 +1,24 @@
...
+        global: &DmabufGlobal,
+        dmabuf: Dmabuf,
+    ) -> Result<Option<D>, ImportError> {
+        todo!("self.renderer.import_dmabuf(&dmabuf, None) ???");
+        Err(ImportError::Failed)
+    }
+
+    fn buffer_created(&mut self, buffer: WlBuffer, result: Option<D>) {
+        todo!()
+    }
```
ABeltramo: I've added the trait `DrmHandler` but I'm a bit lost on how to actually implement this. I was trying to 
```rust
self.renderer.import_dmabuf(&dmabuf, None)
```
but I'm not sure what should I return here.. I've tried `Option<DrmNode>` at first but then it was complaining that
> the trait bound `comp::State: DrmHandler<comp::State>` is not satisfied

I like this approach because everything else in the wl_drm implementation stay the same..

wayland-display-core/src/comp/mod.rs
```diff
@@ -331,16 +342,16 @@ pub(crate) fn init(
...
-                        let toplevel = window.toplevel();
+                        let toplevel = window.toplevel().unwrap();
                         let max_size = Rectangle::from_loc_and_size(
                             (0, 0),
                             with_states(toplevel.wl_surface(), |states| {
                                 states
                                     .data_map
                                     .get::<XdgToplevelSurfaceData>()
-                                    .map(|attrs| attrs.lock().unwrap().max_size)
+                                    .map(|attrs| todo!("Used to be attrs.lock().unwrap().max_size but max_size is now not present!"))
```
ABeltramo: `max_size` is now missing, what am I missing here?

wayland-display-core/src/comp/mod.rs
```diff
@@ -476,27 +487,28 @@ pub(crate) fn init(
...
-                Interest::READ,
-                Mode::Level,
-            ),
-            |_, _, data| {
-                data.display.dispatch_clients(&mut data.state).unwrap();
-                Ok(PostAction::Continue)
-            },
-        )
-        .expect("Failed to init wayland server source");
+    let display_fd = todo!("display.backend().poll_fd().as_raw_fd() -> the trait `AsFd` is not implemented for `i32`");
```
ABeltramo: I've tried a few variations using `as_fd()` but it errors out

wayland-display-core/src/comp/focus.rs
```diff
@@ -117,21 +120,84 @@ impl PointerTarget<super::State> for FocusTarget {
...
             FocusTarget::Wayland(w) => w.axis(seat, data, frame),
             FocusTarget::Popup(p) => p.wl_surface().axis(seat, data, frame),
         }
     }
 
-    fn leave(&self, seat: &Seat<super::State>, data: &mut super::State, serial: Serial, time: u32) {
+    fn frame(&self, seat: &Seat<State>, data: &mut State) {
+        match self {
+            FocusTarget::Wayland(w) => w.frame(seat, data),
+            FocusTarget::Popup(p) => { todo!("the trait `PointerTarget<_>` is not implemented for `&WlSurface`") }
```
ABeltramo: This is still not working..

wayland-display-core/src/comp/mod.rs
Drakulix:
```
states.cached_state.current::<SurfaceCachedState>().max_size
```

wayland-display-core/src/comp/focus.rs
Drakulix: Well there certainly is an implementation for it: https://smithay.github.io/smithay/smithay/input/pointer/trait.PointerTarget.html#impl-PointerTarget%3CD%3E-for-WlSurface

How exactly was your attempt looking? Something like this?
```suggestion
            FocusTarget::Popup(p) => { PointerTarget::frame(p.wl_surface(), seat, data) }
```

wayland-display-core/src/comp/mod.rs
Drakulix: Just staff the whole display into `Generic::new`. Also I don't think we need the `Data` struct anymore and can just use `State` directly. We can call `flush_clients` now on a `DisplayHandle`, which we already have in `State`.

wayland-display-core/src/wayland/handlers/compositor.rs
```diff
@@ -129,7 +128,7 @@ impl CompositorHandler for State {
         }
 
         if let Some(popup) = self.popups.find_popup(surface) {
-            let PopupKind::Xdg(ref popup) = popup;
+            let PopupKind::Xdg(ref popup) = popup else { todo!("refutable pattern in local binding") };
```
Drakulix: Ah right, we also now have input method handling in the popup code. Our compositor doesn't do that, so you can just replace the `todo!` with `unreachable!()`, so it will (most likely) get optimized away.

wayland-display-core/src/wayland/handlers/wl_drm.rs
Drakulix: Oh right, the returned value isn't relevant for this compositor (we don't deal with multiple gpus). So just replace the `Option<D>` with `()` (that is the unit type, aka an empty tuple. Used in place of "Nothing").

`buffer_created` can be an empty implementation and `self.renderer.import_dmabuf(&dmabuf, None).map(|_| ()).map_err(|_| ImportError::Failed)` should work for the `dmabuf_imported` method.

ABeltramo:
The good news is that the compiler is happy and I can finally build and test this; the bad news is that there's nothing in the stream..

The unit test in Wolf where I'm simply importing the buffer and checking it is failing, but when I debug it, it works. So I'm wondering if it's related to your comment about the sync.. I've added this but it might not be enough or not in the right spot?
https://github.com/games-on-whales/gst-wayland-display/blob/63d9e80081d54293718a26590e4ff8c19513b1b5/wayland-display-core/src/comp/mod.rs#L386-L391

Also, I've added a call to go from `info.format()` instead of hardcoding it
https://github.com/games-on-whales/gst-wayland-display/blob/63d9e80081d54293718a26590e4ff8c19513b1b5/wayland-display-core/src/comp/mod.rs#L334-L338
but I'm not sure it's working properly, 
https://github.com/games-on-whales/gst-wayland-display/blob/63d9e80081d54293718a26590e4ff8c19513b1b5/wayland-display-core/src/comp/mod.rs#L300
prints out 
> Requested video format: RGBx .to_fourcc() = 0
so I guess it's going to the hardcoded path anyway.

I'll debug further what's going on with the `GstBuffer` on Wolf side.

ABeltramo:
Or maybe simply this missing bit should be implemented..
https://github.com/games-on-whales/gst-wayland-display/blob/63d9e80081d54293718a26590e4ff8c19513b1b5/wayland-display-core/src/comp/mod.rs#L499-L514

PR #8, opened 2024-11-21
WIP: implement DMABuf support

### LINKED ISSUE #3: Add support for zero-copy pipeline
In our pipeline we are currently doing one unnecessary GPU -> CPU copy in order to expose our raw frame buffer as a
gstreamer buffer. The rationale behind it is explained in
the [docs](https://games-on-whales.github.io/wolf/stable/dev/wayland.html#_gstreamer).   
This will cause some overhead that will eventually be a bottleneck when going with high resolutions and frame rates (on
Discord it has been reported that the current pipeline can't handle 1080p when going over **200 fps**).

There are at least two ways to achieve this in Gstreamer:

- **OpenGL**: exposing the buffer as
  a `GstGLBuffer`, [design docs](https://gstreamer.freedesktop.org/documentation/additional/design/opengl.html)
    - We'll have to share the OpenGL context between our comp and the Gstreamer pipeline.
    - I guess worst case this will end up doing GPU->GPU copy?
- **DMA-BUF**: exposing the buffer as a `GstBuffer` backed by
  a `dmabuf`, [design docs](https://gstreamer.freedesktop.org/documentation/additional/design/dmabuf.html)
    - Cuda doesn't seem to support this directly, (`vaapi` should) it should be possible to import `DMABuf`
      using `glupload -> cudaupload`.
    - I guess worst case this will end up doing GPU->GPU copy?

This needs more research...

## Useful links
 - https://github.com/cmeissl/gst-wlr-screencopy-src/tree/main/src
 - https://github.com/cmeissl/gst_gl_test/blob/main/src/main.rs

### LINKED ISSUE #9: Gamescope window is offset if window decorations are applied
First of all, thanks for this awesome project!

I came across this issue while creating a custom container image for Wolf. Installing Gamescope on Fedora or Ubuntu results in the Gamescope window being offset within gst-wayland-display. Windows are correctly positioned within Gamescope. After removing all libdecor plugins, Gamescope falls back to no window decorations and its window appears in the the correct position.

Gamescope works fine regardless of libdecor plugins if connected to GNOME instead of gst-wayland-display. Programs not using libdecor are positioned correctly within gst-wayland-display. I'm not sure if this issue affects other programs utilizing libdecor.

[games-on-whales/gow](https://github.com/games-on-whales/gow) doesn't seem to be affected because libdecor plugin installation is avoided with `--no-install-recommends`.
 
I created a [repo](https://github.com/eetumos/gow-libdecor) for reproducing this issue. I'm using Fedora 41 with Podman and NVIDIA Container Toolkit. The scripts require modification to work with Docker or Intel/AMD GPUs. Below are screenshots of what I'm experiencing:
 
`./run.sh ubuntu broken gamescope gst`

 
 `./run.sh ubuntu fixed gamescope gst`

ABeltramo:
So that we can finally support zero copy pipelines! There's a [blog post](https://abeltra.me/blog/road-to-zero-copy-in-wolf/) about it too!

Fixes #3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,7 +90,7 @@ dependencies = [
 
 [[package]]
 name = "c-bindings"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "gstreamer",
  "gstreamer-video",
@@ -448,7 +448,7 @@ dependencies = [
 
 [[package]]
 name = "gst-plugin-wayland-display"
-version = "0.2.0"
+version = "0.4.0"
 dependencies = [
  "gst-plugin-version-helper",
  "gstreamer",
@@ -1055,6 +1055,7 @@ dependencies = [
  "appendlist",
  "bitflags 2.4.1",
  "calloop",
+ "cc",
  "cgmath",
  "cursor-icon",
  "downcast-rs",
@@ -1070,6 +1071,7 @@ dependencies = [
  "libc",
  "libloading",
  "once_cell",
+ "pkg-config",
  "profiling",
  "rand",
  "rustix",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,12 +18,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.75"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
-
-[[package]]
 name = "appendlist"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


ABeltramo:
Status update: we can finally exchange DMABuf caps with other plugins and render the wayland output into a DMA buffer to achieve a zero copy pipeline. :rocket: 

## Tested and working pipelines

Simple display output
```
waylanddisplaysrc ! 'video/x-raw(memory:DMABuf),width=1280,height=720,framerate=60/1' 
! glupload ! glimagesink
````

Again, just output the buffer
```
waylanddisplaysrc ! 'video/x-raw(memory:DMABuf),width=1280,height=720,framerate=60/1' 
! glupload ! waylandsink
```

Nvidia HEVC encoding, decoding and displaying.
```
waylanddisplaysrc ! 'video/x-raw(memory:DMABuf),width=1280,height=720,framerate=60/1' 
! glupload ! glcolorconvert ! 'video/x-raw(memory:GLMemory),format=NV12' 
!  nvh265enc ! nvh265dec 
! autovideosink
```
On Nvidia I had to add `GST_GL_API=opengl3` to avoid the warning *OpenGL context could not support PBO download*

mm88ggg:
I’m curious to know the difference in milliseconds of latency between the new version and the previous one.

PR #12, opened 2025-06-13
fix: Only use i915 workaround if first modifier fails

DatCaptainHorse:
It turns out GPUs that are older than DG2 can't handle 4-tiled modifiers afterall, so we'll do a "try, try another and fail" approach instead, to allow both cases to pass for time being.


ABeltramo:
Thanks for this, I've tested it on a CoffeeLake iGPU, and it seems to works correctly now!

PR #13, opened 2025-07-21
Update smithay and rust to latest release

DatCaptainHorse:
Rough update to latest Smithay, does not take advantage of new changes, just made it "compatible".

Marking as draft as there's some oddities I noticed when updating (using smithay's example code as reference), like `pointer_motion_absolute` doing relative mouse update, along other things which seemed odd logically so I removed them for now.


ABeltramo:
I've pushed a simple formatting and minor refactor to add a couple of tests. 

From a quick run everything seems to work, apart from keyboard inputs: I've noticed that the new keyboard `Keysym` has changed the format which results in a mismatch (at least when used with Wolf); see the `keyboard_input` test.

DatCaptainHorse:
> I've pushed a simple formatting and minor refactor to add a couple of tests.
> 
> From a quick run everything seems to work, apart from keyboard inputs: I've noticed that the new keyboard `Keysym` has changed the format which results in a mismatch (at least when used with Wolf); see the `keyboard_input` test.

Yep, xkeysym completely shifted their key values for some reason :shrug: - what I ended up doing on my early-year "messy update" was to just add N to the `keycode` value in `keyboard_input`, which was a workaround. I'll DM you over Discord to communicate faster on this :slightly_smiling_face:

ABeltramo:
Took a little detour to write a micro Wayland client to properly make some integration tests, but I think I've got a better understanding of the low level details of the pointer motion and constraints protocol. The [pointer lock demo](https://mdn.github.io/dom-examples/pointer-lock/) is still a tad odd when running it via Sway, but I think the implementation on our side is as good as it gets.  
Looking forward to your feedback @DatCaptainHorse !

DatCaptainHorse:
@ABeltramo LGTM - tested with this branch on local Nestri build as well to confirm nothing broke, all is good :+1:

Thanks for pushing this PR forward more than I anticipated :tada:

PR #15, opened 2025-08-02
Handle navigation events

anonymix007:


gst-plugin-wayland-display/src/waylandsrc/imp.rs
```diff
@@ -67,104 +67,162 @@ trait EventHandler {
...
+                        // TODO: we need some mapping
+                        //let _ = self.command_tx.send(Command::PointerButton(
+                        //    button as u32,
+                        //    ButtonState::Released
+                        //));
+
+                        return true;
+                    },
+                    NavigationEvent::KeyPress { /*key,*/ .. } => {
+                        // TODO: we need some mapping
```
anonymix007: Not yet sure how to handle this. GStreamer is passing strings which appear to be xkeysyms. I've used `LazyLock<HashMap<String, u32>>` for this in my own plugin, but not yet sure what values are expected here.

gst-plugin-wayland-display/src/waylandsrc/imp.rs
```diff
@@ -67,104 +67,162 @@ trait EventHandler {
...
+                }
+            },
+            gst::EventView::Navigation(n) =>{
+                let navigation_event = gst_video::NavigationEvent::parse(n).unwrap();
+
+                match navigation_event {
+                    NavigationEvent::MouseMove { x, y, .. } => {
+                        let _ = self
+                            .command_tx
+                            .send(Command::PointerMotionAbsolute((x, y).into()));
```
DatCaptainHorse, line 189: I think `Command::PointerMotion` is more approriate, unless navigation MouseMove event provides actual display X,Y coordinates?

anonymix007: It actually does provide the coordinates (relative to the window) as per the documentation: https://gstreamer.freedesktop.org/documentation/video/gstnavigation.html?gi-language=c#gst_navigation_event_parse_mouse_move_event

gst-plugin-wayland-display/src/waylandsrc/imp.rs
ABeltramo: See the `keyboard_scancode_conversion` test, if it's a `XKB` key name you can get the `Keycode` with something like:
```rust
let context = xkb::Context::new(xkb::CONTEXT_NO_FLAGS);
let keymap = xkb::Keymap::new_from_names(&context, "", "", "us", "", None, xkb::COMPILE_NO_FLAGS)
                .unwrap();
let xkb_state = xkb::State::new(&keymap);

let keycode = keymap.key_by_name(key)
```
Not sure if the actual keymap is agreed somewhere else via caps though..

gst-plugin-wayland-display/src/waylandsrc/imp.rs
ABeltramo: There might be some adjustment to do if the window size doesn't match our compositor output size..

ABeltramo:
Thanks for starting this, I have a more general question: **how do we test this?** Which kind of plugins emit navigation events and is there anything that we can simply use without having a full blast WebRTC stream going?  

I've found [navigationtest](https://gstreamer.freedesktop.org/documentation/navigationtest/index.html?gi-language=c#navigationtest) which might be a good way to see if our mouse properly tracks the black squares added, but we need something that generates the events to begin with. Any ideas?

anonymix007:
glimagesink by default will produce these events

gst-plugin-wayland-display/src/waylandsrc/imp.rs
anonymix007: I don't think it is, there doesn't seem to be any checks for these strings.

Here's the example of the mapping that I used: https://github.com/anonymix007/qemu-godisplay/blob/master/keymap.go#L3

ABeltramo: Yeah, if those are the kind of strings in use I don't think there's any way around a custom map..

gst-plugin-wayland-display/src/waylandsrc/imp.rs
anonymix007: Adjustment doesn't seem necessary, glimagesink seems to be already producing the correct coordinates, but cursor seems to be a bit off in Firefox no matter whether window size matches the compositor size or not:
image

Coordinates look reasonable, so could this be related to the cursor being rendered incorrectly?

ABeltramo: I gave it a quick go on my side with https://mousetester.uk/rogueInput.html and it seems that the internal pointer is perfectly tracking. I can see that the rendered cursor looks a bit off, you might be onto something here..

On a side note, if I set `GST_GL_WINDOW=wayland` I get the following warnings:
```bash
# left click
WARN gstwaylanddisplaysrc::waylandsrc::imp: Unknown mouse button pressed: 272
WARN gstwaylanddisplaysrc::waylandsrc::imp: Unknown mouse button released: 272
# right click
WARN gstwaylanddisplaysrc::waylandsrc::imp: Unknown mouse button pressed: 273
WARN gstwaylanddisplaysrc::waylandsrc::imp: Unknown mouse button released: 273
```
(my dev machine is latest KDE Wayland on Arch).
Setting `GST_GL_WINDOW=x11` makes everything working properly.

gst-plugin-wayland-display/src/waylandsrc/imp.rs
```diff
@@ -519,6 +583,230 @@ fn drm_to_gst_format(format: &DrmFormat) -> Option<String> {
...
+        m.insert("KP_Equal", 0x75); // KEY_KPEQUAL
+        m.insert("Pause", 0x77); // KEY_PAUSE
+        m.insert("Meta_L", 0x7d); // KEY_LEFTMETA
+        m.insert("Meta_L", 0x7d); // KEY_LEFTMETA
+        m.insert("Super_L", 0x7d); // KEY_LEFTMETA
+        m.insert("Meta_R", 0x7e); // KEY_RIGHTMETA
+        m.insert("Meta_R", 0x7e); // KEY_RIGHTMETA
+        m.insert("Super_R", 0x7e); // KEY_RIGHTMETA
+        m.insert("Help", 0x8a); // KEY_HELP
+        m.insert("Select", 0x161); // KEY_SELECT
```
ABeltramo: I'd suggest using something like [input-event-codes](https://crates.io/crates/input-event-codes) (or similar library, there are a few apparently) to avoid hard-coding the raw values like that and improve readability.  

ex:
```rust
1 => Some(Command::PointerButton(input_event_codes::BTN_LEFT!(), state)),
```

gst-plugin-wayland-display/src/waylandsrc/imp.rs
anonymix007: Could this be that the hotspot not being handled properly? 

272 and 273 are linux event codes, I'll just also handle them as well. These buttons are definitely platform-specific, so I don't see a sane way to handle consistently these totally unrelated values on different platforms.

gst-plugin-wayland-display/src/waylandsrc/imp.rs
anonymix007: [input-event-codes-sys](https://crates.io/crates/input-event-codes-sys) feels a bit nicer because it autogenerates these from uapi kernel headers at compile time.

gst-plugin-wayland-display/src/waylandsrc/imp.rs
anonymix007: Resolving, this navigation event has absolute coordinates.

anonymix007:
This seems to still work, are there any other blockers for the merge?

P.S. This comment was written in Chromium running in COSMIC/Wayland nested in `waylanddisplaysrc` running in Cinnamon/X11.

ABeltramo:
Just needs another review after the latest changes. Is the issue with the pointer coordinates fixed?   
I'll be away for a few weeks, unless @DatCaptainHorse is available for reviewing you'll have to wait a little longer 😅

anonymix007:
I'm not convinced this issue is related to this PR. 
There are 3 cursors on my monitor right now: one main from Cinnamon (I've added logs and mouse coordinates which element gets look perfectly correct compared to where I move that one), one from COSMIC (and it seems to be where it should) and one from `waylanddisplaysrc` itself. It's only the latter on that seems off. Well, I assume it is from the element itself,  other cursor seems to be changing when e.g. window is resized.

My use case involves running another compositor nested, so I'd want to disable this cursor anyway. I'll probably submit a PR for that as well. One cursor is enough for everyone!

DatCaptainHorse (approved):
I see no issues with the code, and I like the improved EventHandler code 👍

PR #19, opened 2025-10-02
Initial Nix flake

flumf:
This introduces a very basic flake, suitable for building `gst-wayland-display`. It should be relatively low maintenance, as the only moving part is the `cargoHash` that will need to be erased and updated whenever the `Cargo.lock` file is updated.

I'm not very familiar with GitHub Actions, although I expect configuring some kind of runner to ensure this continues to build wouldn't be too difficult: @DatCaptainHorse, would you care to take a look? :P

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -1,0 +1,26 @@
+name: "Nix flake build"
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+    name: Build nix flake
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v13
+
+      - name: Setup Nix cache
+        uses: DeterminateSystems/magic-nix-cache-action@v7
+
+      - name: Check flake
+        run: nix flake check
+
+      - name: Build package
+        run: nix build .#gst-wayland-display --print-build-logs
+
+      - name: Build default package
+        run: nix build --print-build-logs
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ gen
 *.app
 .snapshots/*
 
+# Nix
+/result
+/result-*
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,77 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nix-filter": {
+      "locked": {
+        "lastModified": 1757882181,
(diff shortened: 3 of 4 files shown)


ABeltramo:
Hi @flumf thanks for taking the time to contribute!

Disclaimer: I haven't tested it myself, but I feel that you are missing some stuff from your flake..  Building the package with Cargo isn't the actual goal, we need to produce:

 - A library, `libgstwaylanddisplaysrc.so` so that our plugin can be picked up and used by Gstreamer
 - (optionally) a C header file `libgstwaylanddisplay.h` and a `pkgconfig` file for cross language interop

The way we do that is via [cargo-c](https://crates.io/crates/cargo-c) as specified in our README.

Would you be able to share a few instructions on how to build the library with your flake and then actually use it so that at least a very simple pipeline like 
```
gst-launch-1.0 waylanddisplaysrc ! 'video/x-raw,width=1280,height=720,format=RGBx,framerate=60/1' !  autovideosink
```
would work?

flumf:
Hi!

Currently I've only focused on building the library, and ignored the C bindings. I imagine those could be trivially built as well, if desired, and I may take a look once I've ironed out the initial bugs.

Building the library can be done via `nix build .#`, with the plugin itself being located at `result/lib/libgstwaylanddisplaysrc.so`. From there, following the instructions provided in the repository should be enough.

For testing, I've entered the devShell (`nix develop`) and then run the following:
```
GST_PLUGIN_PATH=result/lib/ gst-launch-1.0 waylanddisplaysrc ! 'video/x-raw,width=1280,height=720,format=RGBx,framerate=60/1' !  autovideosink
```

This does show the plugin loading as expected, although it dies shortly after with the following:
```
ERROR: from element /GstPipeline:pipeline0/GstWaylandDisplaySrc:waylanddisplaysrc0: Internal data stream error.
Additional debug info:
../libs/gst/base/gstbasesrc.c(3187): gst_base_src_loop (): /GstPipeline:pipeline0/GstWaylandDisplaySrc:waylanddisplaysrc0:
streaming stopped, reason not-negotiated (-4)
```

I'm not familiar enough with the internals of this to try and debug it further, though if you have pointers I'd be happy to investigate.

ABeltramo:
That's already pretty much good enough!  

The issue you are encountering is most likely tied to `autovideosink` which may require some more stuff depending on your actual GPU (and/or desktop environment). The easiest pipeline to just test if the library works is probably
```
GST_PLUGIN_PATH=result/lib/ gst-launch-1.0 waylanddisplaysrc render_node=software ! 'video/x-raw,width=1280,height=720,format=RGBx,framerate=60/1' ! fakesink
```
which doesn't require any HW acceleration.

As to how you can make `autovideosink` work, you might need to add `vapostproc` if it's on AMD/Intel as in:
```
GST_PLUGIN_PATH=result/lib/ gst-launch-1.0 waylanddisplaysrc ! 'video/x-raw,width=1280,height=720,format=RGBx,framerate=60/1' ! vapostproc !  autovideosink
```
or something like `cudaupload ! cudaconvert` if using Nvidia as in:
```
GST_PLUGIN_PATH=result/lib/ gst-launch-1.0 waylanddisplaysrc ! 'video/x-raw,width=1280,height=720,format=RGBx,framerate=60/1' ! cudaupload ! cudaconvert !  autovideosink
```
Or you could try replacing `autovideosink` with `glimagesink`.  

If any of the above works, you should be able to see a window with a black background and a mouse for a few seconds, you can then launch applications inside it with something like:
```
WAYLAND_DISPLAY=wayland-1 firefox
```
(assuming you already have a `wayland-0` socket for your desktop)

flumf:
Oh, I see, thank you! Fakesink works as expected, so I'll take a look at getting the C bindings built next.

flumf:
Headers are now installed, along with the libraries for linking against the C bindings and pkg-config files.

PR #20, opened 2025-10-02
[WIP] Producing CUDA buffers 

ABeltramo:
The idea here is that we'll directly produce and push `GSTCudaMemory` buffers, so that we can achieve something like
```
waylanddisplaysrc ! 'video/x-raw(memory:CUDAMemory),width=1920,height=1080,framerate=120/1' ! nvh265enc ! nvh265dec ! autovideosink
```
Basically embedding the whole `glupload ! cudaupload` that we need in order to import our DMA Buffers into CUDA. Ultimately, this should properly fix https://github.com/games-on-whales/wolf/pull/179#issuecomment-3058215348  

## Current status

**It works!**  

The code is a mess and performance is terrible, but I've been iterating over different things for so long that I just need to push something that actually works end to end at displaying a proper image.

We currently incur only in a single GPU->GPU copy when going from CUDA->Gstreamer. I'm not sure if that can be avoided, it feels that we are ultimately doing the same dance that Gstreamer does when importing GL Buffers into CUDA but I would like to dive a bit deeper there.. 

This change will also **help with multi-GPU setups** because we'll be in full control over which Nvidia GPU will be picked and used in the full pipeline.

## TODO

- [x] Take care of the performance issues and clean up the code
- [x] Make the requirement to build and use CUDA optional by properly using [Cargo features](https://doc.rust-lang.org/cargo/reference/features.html)
- [x] See if we can also avoid the GPU->GPU copy
- [ ] Derive the CUDA device ID from the `render_node` param


ABeltramo:
Quick update: code is now in an acceptable state, performance seems to be massively improved, but it feels too good to be true :sweat_smile: 

Screenshot_20251008_200458

I'm going to plug this together in `wolf-ui` and start testing it properly..

PR #21, opened 2025-10-12
feat: CPU optimizations for CUDA-path

DatCaptainHorse:
Profiling with flamegraph showed that some resource allocations were eating plenty of CPU cycles each frame, those are now cached, as the pointer location stays same things work out, but without recreating the resource each time :slightly_smiling_face:

Secondly, `settings.cudabuffer_pool = Some(pool);` was set in `decide_allocation`, which happened later when settings were already communicated to wayland-display-core, so I added extra command to send the buffer pool later on separately, this works out from my testing just before gstreamer pipeline starts rolling.

Thirdly, now that buffer pool is set and available during runtime, I changed `alloc_copy_gst_memory` to return straight buffers instead and separated copy-op to it's own method. Syncing the CUDA stream after unmap was also eating some CPU cycles from simply waiting, as the `CuMemcpy2DAsync` is async, it seems to be fine to let the CUDA runtime handle the unmap later after the async-op is done.

I might've misplaced some comments from original locations, and I'm not sure if `UpdateCUDABufferPool` command is what you'd want so I marked that as TODO :sweat_smile:

Let me know of any issues :pray:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1696,13 +1696,13 @@ dependencies = [
 
 [[package]]
 name = "wayland-backend"
-version = "0.3.10"
+version = "0.3.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe770181423e5fc79d3e2a7f4410b7799d5aab1de4372853de3c6aa13ca24121"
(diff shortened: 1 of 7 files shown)


ABeltramo:
Thank you for taking the time to fine tune this, it's quite amazing the difference in performance that we've achieved by pushing CUDA buffers directly!

PR #23, opened 2025-11-14
feat: Simple clipboard

DatCaptainHorse:
Draft as unsure if can be done "safely" and if `set_data_device_selection` is even needed, should State keep track of the clipboard by itself?

I spent hours just looking for any kind of Smithay example code or documentation, it's **so** lacking, even in Cosmic DE's code I couldn't find anything useful.. So I just write to the FD; what could go wrong? :)

Image below shows clipboard text copied from host to container running gst-wayland-display. Currently limited to text-only, so no images etc.
image


DatCaptainHorse:
Closing as outdated, may revisit later with better implementation unless someone else gets to it before me :slightly_smiling_face:

PR #26, opened 2026-02-25
fix: use proper function to get `VideoFormat` from fourcc

mhebant:
Just a very small fix, as I was also struggling with this and just found a better way

`VideoFormat::from_fourcc` calls the c function `gst_video_format_from_fourcc`
which only works on YUV formats [1] 
Instead use `gst_video_dma_drm_fourcc_to_format` which works on a wider range of formats.

[1] https://gstreamer.freedesktop.org/documentation/video/video-format.html?gi-language=c#gst_video_format_from_fourcc

### DIFF
--- a/gst-plugin-wayland-display/src/waylandsrc/imp.rs
+++ b/gst-plugin-wayland-display/src/waylandsrc/imp.rs
@@ -178,8 +178,8 @@ impl EventHandler for WaylandDisplaySrc {
                     let _ = self.command_tx.send(Command::TouchCancel);
                     return true;
                 }
-            },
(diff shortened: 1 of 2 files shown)


DatCaptainHorse (approved):
Looks good to me, make sure `cargo fmt` is ran if not already, as there's no CI checking that in PRs.

mhebant:
`cargo fmt` formatted some code outside of my patch, just pushed them for good measure

ABeltramo:
Thanks for contributing!  

Re, Cargo format: it was messed up after another PR that was just merged and GitHub Actions aren't running on forks by default. Thanks for pushing that too!

PR #27, opened 2026-04-20
Update running outputs on resize

JBailes:
## What changed
This updates the running-output resize path in `wayland-display-core` so an already active output applies newly negotiated size changes to existing toplevels instead of treating them as a no-op.

## Why
Wolf party mode now resizes seats dynamically when entering split mode. Without this compositor update, running surfaces can ignore the renegotiated size, which leaves the primary seat cropped or stuck on stale geometry.

## Impact
Clients that keep a Wayland output alive while changing negotiated video size now receive the updated fullscreen state and geometry instead of continuing to render against the old size.

## Validation
- `cargo build --release`
- live validation through Wolf party mode on a Moonlight session

### DIFF
--- a/wayland-display-core/src/comp/mod.rs
+++ b/wayland-display-core/src/comp/mod.rs
@@ -262,6 +262,136 @@ impl State {
     }
 }
 
+/// Apply a newly-negotiated `GstVideoInfo` to the compositor state: create or update
+/// the (single) Output's mode, rebuild the damage tracker + allocator, recenter the
+/// pointer, and re-send configure to every mapped toplevel clamped to the new size.
+///
+/// Called from the `Command::VideoInfo` handler and from the test suite. The
+/// `output_already_running` path is what closes the resolution-switching gap --
+/// `space.map_output` must stay one-shot, but everything else is safe and desirable
+/// to re-run on every VideoInfo so connected clients observe the new state.
+pub(crate) fn apply_video_info(
+    state: &mut State,
+    video_info: GstVideoInfo,
+    render_target: &RenderTarget,
+    render_node: Option<DrmNode>,
+) {
+    let output_already_running = state.output.is_some();
+    if output_already_running {
+        tracing::info!("Output already running, updating with newly negotiated video info");
+    }
+    let base_info: VideoInfo = video_info.clone().into();
+    debug!(
+        "Requested video format: {} .to_fourcc() = {}",
+        base_info.format(),
+        base_info.format().to_fourcc()
+    );
+    let size: Size<i32, Physical> = (base_info.width() as i32, base_info.height() as i32).into();
+    let framerate = base_info.fps();
+    let duration = Duration::from_secs_f64(framerate.numer() as f64 / framerate.denom() as f64);
+
+    // init wayland objects
+    let output = state.output.get_or_insert_with(|| {
+        let output = Output::new(
+            "HEADLESS-1".into(),
+            PhysicalProperties {
+                make: "Virtual".into(),
+                model: "Wolf".into(),
+                size: (0, 0).into(),
+                subpixel: Subpixel::Unknown,
+            },
+        );
+        output.create_global::<State>(&state.dh);
+        output
+    });
+    let mode = OutputMode {
+        size: size.into(),
+        refresh: (duration.as_secs_f64() * 1000.0).round() as i32,
+    };
+    output.change_current_state(Some(mode), None, None, None);
+    output.set_preferred(mode);
+    let dtr = OutputDamageTracker::from_output(&output);
+
+    if !output_already_running {
+        state.space.map_output(&output, (0, 0));
+    }
+    state.dtr = Some(dtr);
+    let position = (size.w as f64 / 2.0, size.h as f64 / 2.0).into();
+    state.pointer_location = position;
+    state.pointer_absolute_location = position;
+    state.video_info = Some(video_info.clone().into());
+    match render_target {
+        RenderTarget::Hardware(_) => match video_info {
+            GstVideoInfo::RAW(base_info) => {
+                let allocator = GsGlesbuffer::new(&mut state.renderer, base_info)
+                    .expect("Failed to create GsGlesbuffer");
+                state.output_buffer = Some(GsBufferType::RAW(allocator));
+            }
+            GstVideoInfo::DMA(base_info) => {
+                let allocator = GsDmaBuf::new(render_node.unwrap(), base_info)
+                    .expect("Failed to create GsDmaBuf");
+                state.output_buffer = Some(GsBufferType::DMA(allocator));
+            }
+            #[cfg(feature = "cuda")]
+            GstVideoInfo::CUDA(base_info) => {
+                let egl_display = state
+                    .renderer
+                    .egl_context()
+                    .display()
+                    .get_display_handle()
+                    .handle;
+                let allocator = GsCUDABuf::new(
+                    render_node.unwrap(),
+                    base_info.cuda_context,
+                    base_info.video_info,
+                    Arc::new(Mutex::new(None)),
+                    &egl_display,
+                )
+                .expect("Failed to create GsCUDABuf");
+                state.output_buffer = Some(GsBufferType::CUDA(allocator));
+            }
+        },
+        RenderTarget::Software => {
+            let allocator = GsGlesbuffer::new(&mut state.renderer, base_info.clone())
+                .expect("Failed to create GsGlesbuffer");
+            state.output_buffer = Some(GsBufferType::RAW(allocator));
+        }
+    }
+
+    let new_size = size
+        .to_f64()
+        .to_logical(output.current_scale().fractional_scale())
+        .to_i32_round();
+    for window in state.space.elements() {
+        let toplevel = window.toplevel().unwrap();
+        let max_size = Rectangle::from_size(
+            with_states(toplevel.wl_surface(), |states| {
+                states
+                    .data_map
+                    .get::<XdgToplevelSurfaceData>()
+                    .map(|_attrs| {
+                        states
+                            .cached_state
+                            .get::<SurfaceCachedState>()
+                            .current()
+                            .max_size
+                    })
+            })
+            .unwrap_or(new_size),
+        );
+
+        let new_size = max_size
+            .intersection(Rectangle::from_size(new_size))
+            .map(|rect| rect.size);
+        toplevel.with_pending_state(|state| {
+            state.size = new_size;
+            state.states.set(XdgState::Fullscreen);
+            state.states.set(XdgState::Activated);
+        });
+        toplevel.send_configure();
+    }
+}
+
 pub(crate) fn init(
     command_src: Channel<Command>,
     render: impl Into<RenderTarget>,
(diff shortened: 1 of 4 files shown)


ABeltramo:
It can't be that simple, I refuse to believe that. 😅   
I bet that the something will break on a zero copy pipeline. Have you tested this on Nvidia or just on VAAPI?

JBailes:
> It can't be that simple, I refuse to believe that. 😅 I bet that the something will break on a zero copy pipeline. Have you tested this on Nvidia or just on VAAPI?

Just VAAPI, I don't have an Nvidia card.

ABeltramo:
Ok cool, I'll test it out on my end then. It would be nice if we can also update the tests to check for this resizing if it's actually outputting smaller buffers. Not sure if we have an easy way to exercise that code path though..

JBailes:
> Ok cool, I'll test it out on my end then. It would be nice if we can also update the tests to check for this resizing if it's actually outputting smaller buffers. Not sure if we have an easy way to exercise that code path though..

That was something I struggled with for about an hour than threw my hands in the air and gave up. That's Really Hard.

ABeltramo:
Yeah, looking back at the tests we've only got stuff for the low level buffer stuff but nothing that checks the full Smithay implementation. Now that I think of it, we do have some basic Wayland client unit tests in Wolf: https://github.com/games-on-whales/wolf/blob/7675f0f225b5ffae75806af91e608cff2a8dd59e/tests/platforms/linux/wayland-display.cpp#L59 not sure how complex it would be to test it there, but it might be cool to see if triggering the new event ends up changing the resolution for the Wayland client. This could be a full integration test on Wolf side.

JBailes:
> Yeah, looking back at the tests we've only got stuff for the low level buffer stuff but nothing that checks the full Smithay implementation. Now that I think of it, we do have some basic Wayland client unit tests in Wolf: https://github.com/games-on-whales/wolf/blob/7675f0f225b5ffae75806af91e608cff2a8dd59e/tests/platforms/linux/wayland-display.cpp#L59 not sure how complex it would be to test it there, but it might be cool to see if triggering the new event ends up changing the resolution for the Wayland client. This could be a full integration test on Wolf side.

I already have an integration test harness suite in PR -> https://github.com/games-on-whales/gow/pull/321 Once this gets feedback and merged, I should be able to adapt the pattern into integration tests for Wolf.

wayland-display-core/src/comp/mod.rs
```diff
@@ -261,6 +262,136 @@ impl State {
...
+            .unwrap_or(new_size),
+        );
+
+        let new_size = max_size
+            .intersection(Rectangle::from_size(new_size))
+            .map(|rect| rect.size);
+        toplevel.with_pending_state(|state| {
+            state.size = new_size;
+            state.states.set(XdgState::Fullscreen);
+            state.states.set(XdgState::Activated);
```
ABeltramo, line 389: These are additions compared to the previous code, I guess it doesn't matter much since we support one single window anyway but I was wondering what was the reason behind this change?

wayland-display-core/src/tests/client.rs
```diff
@@ -226,6 +237,55 @@ impl WaylandClient {
...
+        })
+    }
+
+    /// The most recent toplevel configure seen on the first window.
+    pub fn latest_configure(&self) -> Option<Configure> {
+        self.state
+            .windows
+            .first()
+            .and_then(|w| w.configures_received.last().map(|(_, c)| c.clone()))
+    }
```
ABeltramo: These functions aren't referenced anywhere, it seems to be dead code.

wayland-display-core/src/comp/mod.rs
JBailes: It's a defensive change:  every configure that goes out during a live resize should declare the desired xdg states explicitly rather than rely on whatever got set on the initial map.

wayland-display-core/src/tests/client.rs
JBailes: Yep, think I mentioned that previously. Do you want me to remove the code instead of testing it?

ABeltramo: Yes please, no point in keeping extra test code that it's not used..

JBailes: :+1: You got it!

JBailes: Should be fixed.

ABeltramo:
I don't see anything wrong with the implementation, but at the same time I'd like to properly test this in a Gstreamer pipeline, and I was thinking how to do that..  

Something that I quite like when developing `gst-wayland-display` is to just run a quick pipeline via `gst-launch-1.0`; ex:
```bash
gst-launch-1.0 waylanddisplaysrc render_node=...nvidia-gpu-here... ! 'video/x-raw(memory:CUDAMemory),width=1920,height=1080,framerate=60/1' ! cudaconvert ! cudadownload ! autovideosink
```
or on AMD/Intel something like:
```bash
gst-launch-1.0 waylanddisplaysrc render_node=...amd-gpu-here... ! 'video/x-raw(memory:DMABuf),width=1920,height=1080,framerate=60/1' ! vapostproc ! autovideosink
```
Then, on a second CLI you can easily run any app like
```bash
WAYLAND_DISPLAY=wayland-1 gamescope -- firefox
```

In our implementation we have support for the `Navigation` events and so mouse and keyboard events will be automatically forwarded from `autovideosink` to our `gst-wayland-comp`. **It would be nice if we could easily test out the resolution change this way**! We could then easily add encoders in the mix like `nvh265enc ! nvh265dec ! cudaconvert ! cudadownload ! autovideosink` to see if any of that breaks too.  

Now, from a quick glance, it seems that the underlying video sinks like [glimagesink](https://gstreamer.freedesktop.org/documentation/opengl/glimagesink.html?gi-language=c) deal with the video scaling internally and don't expose window resizing as events. Something that might help here that I had never tried to use myself could be [gst-validate-launcher](https://gstreamer.freedesktop.org/documentation/gst-devtools/gst-validate-launcher.html?gi-language=c) where we basically write a scenario that after a few seconds tries to change the buffer size and trigger a `Reconfigure` event (I think this is the right type of event but we should check), or even just directly a `CustomUpstream` event that we then add support for.  

We have to investigate this anyway because I think **we would need a clean way to signal to the pipeline that the resolution has changed so that all components** (mainly thinking about encoders here) **resize accordingly**, so I think it's worth to take a look at this before merging this.

JBailes:
Good questions, Abe, poking into it some myself.

JBailes:
@ABeltramo Results added to the PR, including specific manual testing steps for someone with an nVidia GPU (Which I don't have).

ABeltramo:
Have you actually tested this? It looks to me that `gst-validate-1.0` doesn't show a window and even if that worked fine changing the caps like that isn't the right way of achieving this AFAIU.  
See for example the official docs about [negotiation](https://gstreamer.freedesktop.org/documentation/additional/design/negotiation.html?gi-language=c#pushmode-negotiation) or [bufferpool](https://gstreamer.freedesktop.org/documentation/additional/design/bufferpool.html?gi-language=c#renegotiation) you really have to send and add support for `RECONFIGURE` events.

JBailes:
@ABeltramo I did a quick and dirty test. I'm engineering a much larger/more correct test now and testing across Intel and AMD.

JBailes:
@ABeltramo Three "Output already running, updating with newly negotiated video info" logs fired at exactly the wall-clock moments each caps flip propagated through - 5s, 10s, 15s. With a real Wayland client producing frames, the pipeline clock advanced, RECONFIGURE propagated, set_caps was called three times, and PR 27's new "updating" branch was hit three times. On the stable plugin this line would say "ignoring" - which is the whole point of the fix. Will update with more results.

JBailes:
@ABeltramo Hardware DRM path still fails at negotiate()
Testing with a full Moonlight setup and Wolf using the 27 PRs Wayland resulted in a change of resolution and gave the mentioned logs previously.

These results were across Intel and AMD.

Let me know what more you would want.

ABeltramo:
> Three "Output already running, updating with newly negotiated video info" logs fired at exactly the wall-clock moments each caps flip propagated through — 5s, 10s, 15s

Can you expand a little on how exactly are you running this test? If I run `gst-validate-1.0` like described in your README it completes in less than a second and doesn't show anything.

> Hardware DRM path still fails at negotiate()

So it's not working, are you testing this just with SW..?

> Testing with a full Moonlight setup and Wolf using the 27 PRs Wayland resulted in a change of resolution and gave the mentioned logs previously.

Do you mean `"Output already running, updating with newly negotiated video info"`? So is it working..?

JBailes:
Good callout on 1, @ABeltramo , you are right, the README isn't accurate with the larger test (I'm preparing to push it, it's not quite there yet)

Pre-reqs: apt-get install -y python3-gi python3-gst-1.0 weston

  Harness (pr27-midstream.py):

  import gi, os, sys, time, subprocess, threading
  gi.require_version("Gst", "1.0")
  from gi.repository import Gst, GLib

  Gst.init(None)

  CAPS = [
      (5,  "video/x-raw,width=1280,height=720,format=RGBx,framerate=60/1"),
      (10, "video/x-raw,width=2560,height=1440,format=RGBx,framerate=60/1"),
      (15, "video/x-raw,width=1920,height=1080,format=RGBx,framerate=60/1"),
  ]

  pipeline = Gst.parse_launch(
      "waylanddisplaysrc render_node=software name=src ! "
      "capsfilter name=res caps=video/x-raw,width=1920,height=1080,format=RGBx,framerate=60/1 ! "
      "videoconvert ! fakesink sync=false"
  )
  loop = GLib.MainLoop()

  def on_msg(bus, msg):
      if msg.type == Gst.MessageType.ERROR:
          e, d = msg.parse_error()
          print(f"[pipeline ERROR] {e.message} :: {d}", flush=True); loop.quit()
      elif msg.type == Gst.MessageType.STATE_CHANGED and msg.src == pipeline:
          old, new, _ = msg.parse_state_changed()
          print(f"[pipeline state] {old.value_nick} -> {new.value_nick}", flush=True)

  bus = pipeline.get_bus(); bus.add_signal_watch(); bus.connect("message", on_msg)

  def set_caps(caps_str):
      pipeline.get_by_name("res").set_property("caps", Gst.Caps.from_string(caps_str))
      print(f"[harness t={time.time()-t0:.1f}s] SET caps => {caps_str}", flush=True)
      return False

  def start_client():
      for _ in range(40):
          socks = [s for s in os.listdir(os.environ["XDG_RUNTIME_DIR"])
                   if s.startswith("wayland-") and not s.endswith(".lock")]
          if socks:
              env = os.environ.copy(); env["WAYLAND_DISPLAY"] = socks[0]
              subprocess.Popen(["weston-simple-egl"], env=env,
                               stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
              print(f"[harness] compositor socket {socks[0]} (attached weston-simple-egl)", flush=True)
              return
          time.sleep(0.25)

  t0 = time.time()
  pipeline.set_state(Gst.State.PLAYING)
  threading.Thread(target=start_client, daemon=True).start()

  for secs, caps in CAPS:
      GLib.timeout_add_seconds(secs, set_caps, caps)

  def finish():
      pipeline.set_state(Gst.State.NULL); loop.quit(); return False
  GLib.timeout_add_seconds(18, finish)
  loop.run()

  Invocation (assuming the PR 27 plugin is staged at /tmp/pr27/libgstwaylanddisplaysrc.so):

  GST_PLUGIN_PATH=/tmp/pr27:/usr/local/lib/x86_64-linux-gnu/gstreamer-1.0 \
  RUST_LOG=info \
  XDG_RUNTIME_DIR=/run/user/$(id -u) \
  python3 pr27-midstream.py

JBailes:
> So it's not working, are you testing this just with SW..?

SW and HW. I am trying to keep this as much to testing harnesses as we can, anything I do manually is much more rough. Note the above script *should* be able to switch between software and hardware. (Works on my machine caveat, prepping as part of the larger PR)

> Do you mean "Output already running, updating with newly negotiated video info"? So is it working..?

If you want to phrase it as such, this is being tested manually with Moonlight and the full Wolf path as well. I *did* just find an Intel bug thanks to this.

ABeltramo:
I'm a still confused on the results and tests, but I'll give you time to work on this and push what you've got..  
Don't rush stuff, take you time, run the tests, you don't have to reply in less than 5 minutes this stuff takes time to do it properly :sweat_smile:

JBailes:
Okay. Now that I've got the PR updated with the most up to date testing, let me sit down and explain a bit better:

Testing followed 3 major lanes:
1) Testing 3 cases with AMD (7900XTX) and Intel (14700k IGPU) hardware running the full Wolf stack with a PR27-based wayland display and standard moonlight.
This passed. Including properly resizing the screen. Also, software worked for this, it was just much slower, but all 3 cases passed. These were manual testing, I tried to create a script to do it automatically, but this was just a frustrating exercise of chasing my tail. Human eyes are required for this part right now, but I'll figure out how to automate this in the future.
2) Testing with integration tests
See PR. The integration tests are as close as we can get to raw hardware on CI, this is intentional.
3) Script testing working against the wayland code directly.
See PR and the python script.

JBailes:
Found another AMD-specific bug in manual testing, fixing.

JBailes:
I take it back, it was a party mode bug that needs to be fixed in the party mode PR.

JBailes:
image

Manual testing.

https://github.com/games-on-whales/wolf/pull/396 is what is needed to manually test. You'll have to build a wayland off this branch and than use that. Trigger party mode for the resize. I'm just finding tons of 396 bugs to fix at this point, no further 27 bugs yet.

JBailes:
image

At this point, it is looking like Wolf fixes are left, not fixes for here.

JBailes:
@ABeltramo Okay, 100% confidence we're done with this.

It's been pretty exhaustively tested with Party Mode. TBH, the only change I had to make was in docs/the test harness for your asks, the core code itself has been good to go. There's some Wolf fixes that I need to make, but that's independent of this PR.

ABeltramo:
I'll test out on Nvidia as soon as I have a bit of time. No need to commit the python test, it's just throwaway code that shouldn't be part of the PR..

JBailes:
image

With the set of wolf fixes I needed to make. Python script is deleted.

PR #29, opened 2026-05-11
Raise Wayland display max buffer size

JBailes:
## Summary
- Update Wayland crates and enable `wayland-server` with `libwayland_1_23`.
- Set the compositor display default max buffer size to 10 MiB before state initialization.
- Mirrors the relevant behavior from https://github.com/linuxserver/pixelflux/commit/ffc2d4daef02726857174fd3a6b158a53ac09c5c.

## Verification
Fresh CTs were created on the Proxmox host for verification. CT120 was not used.

- CT900 Intel RPL-S, host render node `/dev/dri/renderD129`: `test_dmabuf` passed.
- CT900 Intel: `cargo test --locked -p wayland-display-core --lib -- --skip test_dmabuf` passed, 12 passed.
- CT901 AMD Radeon RX 7900 XTX, host render node `/dev/dri/renderD130`: `test_dmabuf` passed.
- CT902 NVIDIA GeForce RTX 5080, host render node `/dev/dri/renderD128`: installed matching NVIDIA 595.71.05 userspace-only from the NVIDIA runfile, mapped `/dev/nvidia*` and `/dev/nvidia-caps`, and verified `nvidia-smi` plus `eglinfo -B` report NVIDIA EGL on 595.71.05.
- CT902 NVIDIA: non-dmabuf library suite passed with `XDG_RUNTIME_DIR=/tmp/gwd-xdg`, 12 passed.
- CT900: `cargo check --workspace --all-targets` passed, and feature tree confirmed `wayland-server/libwayland_1_23` enables `wayland-backend/libwayland_server_1_23`.

## NVIDIA AB24 Linear note
`utils::allocator::tests::test_dmabuf` still fails on NVIDIA after the CT was converted into a valid NVIDIA EGL environment. The failure is:

- `DrmFourcc(AB24), Modifier: Linear`
- `GL_INVALID_OPERATION ... EGLImage not supported`
- panic at `wayland-display-core/src/utils/allocator/mod.rs:682`, `assertion failed: bind_result.is_ok()`

This was also reproduced against a clean `HEAD` baseline before this change, using the same CT902 NVIDIA environment and `GBM_BACKEND=nvidia-drm`. Therefore the NVIDIA AB24 Linear import issue predates this PR and is not introduced by the Wayland max-buffer change.

CT900, CT901, and CT902 are left running for inspection.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -354,9 +354,9 @@ dependencies = [
 
 [[package]]
 name = "dlib"
-version = "0.5.2"
+version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "330c60081dcc4c72131f8eb70510f1ac07223e5d4163db481a04a0befcffa412"
+checksum = "ab8ecd87370524b461f8557c119c405552c396ed91fc0a8eec68679eab26f94a"
 dependencies = [
  "libloading",
 ]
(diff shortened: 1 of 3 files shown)


ABeltramo:
Just posting this here so that we don't forget: we should bump up the SHA that we bundle in our Dockerfile in Wolf so that this will be picked up.

PR #30, opened 2026-06-01
fix: devcontainer podman build

barne856:
There were a few issues I ran into building this devcontainer with podman. First: `cargo-c` has bumped its minimum required rust version to `1.93` https://github.com/lu-zero/cargo-c/blob/e7a5919aaba93e3bcff0932fc830cec654aeca36/Cargo.toml#L12-L13 causing the install in the Dockerfile to fail. Second: the `--runtime=nvidia` is legacy and the new recommend method of accessing the GPU inside the container is through the CDI https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html which does not require `privileged` to be true or `hostRequirements` to be set.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -27,7 +27,7 @@ RUN apt-get update -y \
                libwayland-dev libwayland-server0 libinput-dev libxkbcommon-dev libgbm-dev \
                librust-gstreamer-allocators-dev
 
-ARG RUST_VERSION=1.89.0
+ARG RUST_VERSION=1.96.0
 ENV RUST_VERSION=$RUST_VERSION
 ENV PATH="$HOME/.cargo/bin:${PATH}"
 
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -27,8 +27,6 @@ RUN apt-get update -y \
                libwayland-dev libwayland-server0 libinput-dev libxkbcommon-dev libgbm-dev \
                librust-gstreamer-allocators-dev
 
-ARG RUST_VERSION=1.96.0
-ENV RUST_VERSION=$RUST_VERSION
 ENV PATH="$HOME/.cargo/bin:${PATH}"
 
 USER $UNAME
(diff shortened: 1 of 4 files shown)


.devcontainer/default/devcontainer.json
```diff
@@ -5,11 +5,7 @@
     "dockerfile": "Dockerfile"
   },
   "features": {},
-  "runArgs": ["--runtime=nvidia"],
-  "hostRequirements": {
-    "gpu": true
-  },
-  "privileged": true,
```
ABeltramo, line 12: `privileged` here wasn't for accessing Nvidia GPUs, but it was mostly to get access to `/dev/` so that we get virtual devices support and `/dev/dri/` for Intel/AMD GPUs.  

Since this is a dev container I think it's fine to keep it, but I'm open to alternatives.

.devcontainer/Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN apt-get update -y \
                libwayland-dev libwayland-server0 libinput-dev libxkbcommon-dev libgbm-dev \
                librust-gstreamer-allocators-dev
 
-ARG RUST_VERSION=1.89.0
```
ABeltramo, line 30: This will inevitably become old again, I wonder if we should just pull the latest stable version since this is a dev container?

ABeltramo (commented):
Thanks for taking the time to contribute! I've added a couple comments on the changes; see below

.devcontainer/default/devcontainer.json
barne856: Thanks for the context, I added `privileged` back into the devcontainer. One thing to consider, if the intent is to support non-nvidia systems as well as nvidia systems then `--device nvidia.com/gpu=all` will not work for them, in that case we'd need 2 devcontainers one with and one without the flag. VS Code supports this by creating subdirectories under `.devcontainer/` I made a commit with this if you want to go that route.

.devcontainer/Dockerfile
barne856: agreed, updated to install stable toolchain

.devcontainer/default/devcontainer.json
ABeltramo: That's interesting, how would it decide which devcontainer to use? Or is there a UI in vscode that lets you pick one of the available devcontainers?

barne856: Yes when building the container it will prompt you to choose one if there are multiple available.

ABeltramo:
Thanks for all the info and for working on this, looks good to me!

PR #31, opened 2026-06-08
feat(comp): explicit sync (linux-drm-syncobj-v1)

JBailes:
> Consider this an idea/draft PR — it's one of the issues I've been hitting with Blackwell + the 610 drivers, and I was teaching myself Rust along the way. Happy to iterate.

## Problem

Without explicit sync, the compositor relies on **implicit** dmabuf fences to know when a client has finished rendering into a buffer. NVIDIA's driver doesn't do implicit sync, so the compositor can sample a half-rendered buffer — the flicker/tearing/jerk class of artifacts. I can reproduce this on NVIDIA where it doesn't happen on AMD (e.g. Bravely Default: Flying Fairy tears/jerks when moving quickly around maps). Vulkan is explicit by design and NVIDIA needs it; NVIDIA's new dma-buf platform lib ([egl-wayland2](https://github.com/NVIDIA/egl-wayland2)) calls out `linux-drm-syncobj-v1` as required "for full functionality… without it you may get reduced performance and out-of-order frames."

## What this does

Implements `wp_linux_drm_syncobj_manager_v1` (explicit sync) in the compositor, mirroring the smithay anvil integration:

- On a hardware render target, open the render node as a `DrmDeviceFd` and, when it supports syncobj eventfd, advertise a `DrmSyncobjState` global (`comp/mod.rs`); otherwise log and disable.
- A `new_surface` pre-commit hook installs an **acquire-point blocker**: a buffer is presented only once its explicit-sync fence signals (`handlers/compositor.rs`).
- **Implicit-fence fallback**: when a client *isn't* using explicit sync, fall back to blocking the commit on the dmabuf's implicit read-fence (`Interest::READ`), matching anvil. This means dmabuf clients get correct commit-time sync whether or not they speak `linux-drm-syncobj-v1`.
- Release points are signalled automatically by smithay (`DrmSyncobjCachedState::merge_into` / destruction hook) — no manual signalling needed.

Empirically this removes the tearing/jerking for me on the NVIDIA 610 drivers.

## Caveats

- This depends on what the target compositor reports in our nested setup — sway/gamescope have to support the protocol for a client to actually use it. `WAYLAND_DEBUG=client|server` is useful to watch the negotiation.
- 610/Blackwell has a few other rough edges I'm still looking at (XDG behaviour, block-linear render buffers — the latter tracked in games-on-whales/wolf#417). Those are intentionally **not** in this PR anymore; it's now purely explicit sync.

## Refs

- games-on-whales/wolf#417 (block-linear, separate)
- https://github.com/NVIDIA/egl-wayland2

### DIFF
--- a/wayland-display-core/src/comp/mod.rs
+++ b/wayland-display-core/src/comp/mod.rs
@@ -207,28 +207,34 @@ impl State {
         };
 
         let drm_syncobj_state = if let RenderTarget::Hardware(node) = render_target {
-            match std::fs::OpenOptions::new()
-                .read(true)
-                .write(true)
-                .open(node.dev_path().expect("Failed to determine DrmNode path?"))
-            {
-                Ok(file) => {
-                    let device_fd = DrmDeviceFd::new(DeviceFd::from(OwnedFd::from(file)));
-                    if supports_syncobj_eventfd(&device_fd) {
-                        tracing::info!("Enabling explicit sync (linux-drm-syncobj-v1)");
-                        Some(DrmSyncobjState::new::<State>(&dh, device_fd))
-                    } else {
+            match node.dev_path() {
+                Some(path) => match std::fs::OpenOptions::new()
+                    .read(true)
+                    .write(true)
+                    .open(&path)
+                {
+                    Ok(file) => {
+                        let device_fd = DrmDeviceFd::new(DeviceFd::from(OwnedFd::from(file)));
+                        if supports_syncobj_eventfd(&device_fd) {
+                            tracing::info!("Enabling explicit sync (linux-drm-syncobj-v1)");
+                            Some(DrmSyncobjState::new::<State>(&dh, device_fd))
+                        } else {
+                            tracing::warn!(
+                                "DRM device does not support syncobj eventfd; explicit sync disabled"
+                            );
+                            None
+                        }
+                    }
+                    Err(err) => {
                         tracing::warn!(
-                            "DRM device does not support syncobj eventfd; explicit sync disabled"
+                            ?err,
+                            "Failed to open render node for syncobj; explicit sync disabled"
                         );
                         None
                     }
-                }
-                Err(err) => {
-                    tracing::warn!(
-                        ?err,
-                        "Failed to open render node for syncobj; explicit sync disabled"
-                    );
+                },
+                None => {
+                    tracing::warn!("Render node has no device path; explicit sync disabled");
                     None
                 }
             }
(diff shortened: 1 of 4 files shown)


ABeltramo:
What's the actual issue though?

JBailes:
> What's the actual issue though?

Oof. You can see how long I've been pounding my head out with 610/blackwell drivers that I forgot to include this basic thing in the description. Also keep in mind I was teaching myself Rust with this PR.

Explicit sync (linux-drm-syncobj-v1) is about buffer sync between client and compositor. Without it the compositor leans on implicit dmabuf fences to know when a client's render finished. NVIDIA's driver doesn't do implicit sync, so the compositor can sample a half-rendered buffer (the flicker/tearing/glitch class). Explicit sync swaps that for acquire/release timeline points passed over the protocol: client says "ready when this fence signals," compositor signals "done" when it's finished reading. Vulkan's explicit by design and NVIDIA needs it, so it's helpful with nvidia's 610 drivers.

Please note that this should have really been a draft PR and is more of an idea PR. There's also another area I've identified with 610 that also needs some focus, that's the XDG area. 610 had three major...changes? Regressions? that we care about. The explicit sync, XDG, and block linear.

ABeltramo:
> 610 had three major...changes? Regressions? that we care about. The explicit sync, XDG, and block linear

Have you got a link to that? I'd like to read more about it..

Be aware that with our stack we also depend on what the target compositor (sway/gamescope) supports reporting in this weird setup. It wouldn't surprise me if they don't implement the protocol for explicit sync. You can see the Wayland protocol stuff in real-time with an env variable `WAYLAND_DEBUG` which can be `client` or `server`  IIRC. There are also TUI/GUI that would help you visualise what's going on if you want to dig and understand more..

JBailes:
NVIDIA/egl-wayland2 https://github.com/NVIDIA/egl-wayland2 the new dma-buf platform lib: "only block-linear DMA buffer layouts are supported on NVIDIA Wayland," requires the dma-buf protocols, and needs linux-drm-syncobj-v1 "for full functionality… without it you may get reduced performance and out-of-order frames"

JBailes:
@ABeltramo To put it simpler, I can see tearing and screen jerkiness on nvidia drivers that don't exist on AMD's.

As one example, with Bravely Default: Flying Fairy, I can see a bit of tearing/jerking when I am moving quickly around maps.

To the best of my understanding, this is due to explciit sync.. Running this code gives me no more jerking/tearing.

JBailes:
@ABeltramo Did a big rewrite, see what you think.

ABeltramo (approved):
I've taken a look at the implementation and compared with the upstream Smithay Anvil implementation which I think we are tracking properly. I can see that we are correctly checking for `supports_syncobj_eventfd` so it looks like this would work for non Nvidia GPUs; I'm just not sure about the SW renderer.

All in all I think this is a great addition and cleanly done. I'll merge this and properly test it out in a Wolf PR with different GPUs. Good job!

PR #33, opened 2026-06-11
refactor: stop advertising legacy wl_drm when dmabuf v4 feedback is available

shrinedogg:
## What this changes

`wl_drm` is a legacy Mesa protocol, superseded by `zwp_linux_dmabuf_v1` v4 feedback for render-device discovery — Mutter and KWin have already dropped it entirely. Today the compositor advertises **both** `wl_drm` and dmabuf v4 default feedback at the same time, which is redundant.

This PR stops advertising `wl_drm` when the dmabuf v4 default-feedback global is created successfully:

* skip `create_drm_global` when v4 default feedback is available;
* keep `wl_drm` on the dmabuf **v3** fallback path, where clients still need it for device discovery;
* add an opt-in escape hatch for legacy clients — `GST_WAYLAND_DISPLAY_ADVERTISE_WL_DRM=1` restores the previous behavior.

## Scope / history (please read)

I originally opened this believing it fixed a nested-wlroots (sway) startup abort — `assert(wl->drm_render_name == NULL)` at `backend/wayland/backend.c:311`. Per @JBailes's review, **that justification does not hold up**: wlroots 0.19 already handles `wl_drm` and dmabuf-v4 feedback coexisting (it destroys the `wl_drm` global and resets `drm_render_name` to `NULL` before reading the device from the v4 `main_device`), and the logs I attached don't actually contain that assert. Thanks for the careful read.

So, taking @JBailes's suggestion to split this up:

* **This PR no longer claims to fix that crash.** It stands on its own as a cleanup that matches what Mutter/KWin already did and removes a redundant legacy global, with an env opt-out for old clients.
* The problem actually visible in my logs is **separate** and unrelated to `wl_drm`: inside the session pod, wlroots can't resolve the `main_device` dev_id advertised in the v4 feedback (`render/wlr_renderer.c:249 drmGetDevices2 failed`, `backend/wayland/backend.c:166 drmGetDeviceFromDevId failed`), and sway later dies on `output.c:525 Failed to commit output WL-1` (no outputs connected). That's a DRM-node-in-namespace issue; I'll investigate and raise it on its own rather than bundling it here.

## Testing

Build only (`cargo check`) — this is purely a protocol-advertisement change and I'm intentionally **not** attaching a crash-fix claim to it. Happy to add the env hatch behind a different name or default if you'd prefer.

### DIFF
--- a/wayland-display-core/src/comp/mod.rs
+++ b/wayland-display-core/src/comp/mod.rs
@@ -107,7 +107,7 @@ pub struct State {
     pub(crate) output_buffer: Option<GsBufferType>,
     render_node: Option<DrmNode>,
     pub renderer: GlesRenderer,
-    dmabuf_global: Option<(DmabufGlobal, GlobalId)>,
+    dmabuf_global: Option<(DmabufGlobal, Option<GlobalId>)>,
     last_render: Option<Instant>,
 
     // management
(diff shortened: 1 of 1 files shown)


ABeltramo:
Thanks for taking the time to contribute.  

It's the first time that someone reports Sway aborting on startup, would you be able to share the full Wolf logs and a bit more info about this? Is it running a custom app image or one of the defaults from GOW? I'd like to reproduce this if possible..

shrinedogg:
Hey ABeltramo; happy to contribue and thank you for the groundwork on Fenrir, I love it!

So, there was a handful of things I had to work on and spend tokens on to get this going. It's probably easier to just take a look here: https://github.com/shrinedogg/biggs.dog#-dreamcast-game-streaming-stack

Here are some logs for all the "alex-steam" pods I've had since June 10th when I started this adventure with Wolf. These errors were on whatever the default Wolf image that the Fenrir release uses. I have since switched to the ones I have worked on.
 
[Logs-06_12_2026, 07_05_43 PM.txt](https://github.com/user-attachments/files/28887300/Logs-06_12_2026.07_05_43.PM.txt)

If you do a find for "wayland" that might help shed light on what the errors are.

JBailes:
I went through both the wlroots 0.19 source and our compositor and I don't think the root cause here is right.

Yeah, we do advertise both wl_drm and dmabuf-v4 feedback together (comp/mod.rs:184-202). But wlroots 0.19 actually expects that, when it sees v4 default feedback it destroys the wl_drm global and resets drm_render_name back to NULL before pulling the name from the feedback's main_device (backend/wayland/backend.c:643-672). So the asserts at :311 and :183 are guarding against the name being set twice from the same source, not against the two globals coexisting. Having both up is the handled case, not the bug.

Also the attached log doesn't actually show this, there's no backend.c:311, no assert, no sway in it. It's a gamescope pod aborting on SIGTERM at session stop (See line 88 on the log, "init-gamescope.sh: executing..."), plus unrelated renderD128/dbus/nvrtc noise.

No objection to gating wl_drm behind the v4 check, Mutter/KWin dropped it too and the env escape hatch is fine, but the justification doesn't line up, and if a nested compositor really is asserting I don't think this is what's fixing it. Could you share a log that actually shows the :311 abort from a nested wlroots client?

shrinedogg:
My apologies, those logs are missing many lines, I should have double checked. I tried to export them from Grafana and it didn't do a great job, but here's some offending lines:

image

Here's some better logs with info filtered out, but still not all of them. This set of logs covers the afternoon of June 10th (around 1300-1700). I don't think they show exactly what your looking for, but this did seem to resolve some Sway issues for me. I did more to get it working with the controller as well, there we some window focus issues in Steam when running an application.

[Logs-06_13_2026, 11_13_27 AM.txt](https://github.com/user-attachments/files/28909459/Logs-06_13_2026.11_13_27.AM.txt)

I honestly went pretty hard on this over this week to get it to where it is running on my Talos node. I've updated my README here: https://github.com/shrinedogg/biggs.dog#patched-fork--engineering-notes with the changes and intend to contribute them all this weekend.

JBailes:
I've had a quick read through the logs, but I'm still in a position where I'm scratching my head as to why the proposed code would fix what you are stating. As I mentioned, wlroots 0.19 already expects this, it shouldn't cause this behavior in your client. These logs still don't show what the PR is fixing. There's no backend.c:311, no assert, no abort anywhere in the file, just three ERROR lines on each relaunch:

[backend/wayland/backend.c:261] drmGetDevices2 failed: No such file or directory
[backend/wayland/backend.c:166] drmGetDeviceFromDevId failed: No such file or directory
[render/wlr_renderer.c:249]     drmGetDevices2 failed: No such file or directory

Those aren't fatal, sway keeps going and only dies later on something unrelated (output.c:525 Failed to commit output WL-1 -> no outputs connected).

And the third line is the giveaway: wlr_renderer.c:249 is the dmabuf/renderer path resolving the main_device dev_id we advertise in the v4 feedback (comp/mod.rs:182), nothing to do with wl_drm. wlroots can't map that dev_id to a node inside the nested client, the DRM node just isn't there. Gating wl_drm won't touch that line, so if anything the log argues against the fix.

No problem dropping wl_drm as a cleanup, Mutter/KWin did the same and the env hatch is fine. But I can't take it as a crash fix on this evidence. If a nested client really is hitting the :311 assert, can you grab a log that actually contains that line? Otherwise I'd rather split this into the wl_drm cleanup (no crash claim) and a separate look at the main_device not resolving in the client namespace.

shrinedogg:
Hey JBailes, thanks for having a look and I agree with your conclusion. I can't find any logs on right now to prove otherwise, so I am not sure how I wound up on that path now. 😦 

 I've updated the OG PR for now, or I can close and resubmit new ones, up to you!

JBailes:
After  understanding more about it, I'm inclined to not merge this unless the benefit for the merge is better explained. It was originally reported as a crash bug, and now the impetus behind merging it seems to be because if's what kwin/mutter did.

PR #34, opened 2026-06-17
feat(waylandsrc): add await-listener to defer caps negotiation until a consumer connects

JBailes:
## Problem

When `waylanddisplaysrc` feeds a decoupled consumer through an `interpipesink` (as Wolf does for streaming), the producer pipeline starts and **fixes its output format before any consumer is listening**. With a generic producer cap such as `video/x-raw(memory:DMABuf)`, the compositor settles on the first advertised drm-format, which on both AMD and Intel is a 64-bit float format (`AB4H`). Downstream converters like `vapostproc` can't ingest it, so negotiation fails with `not-negotiated` once the client actually connects. Live renegotiation can't recover it: the in-flight buffer and the producer capsfilter are already anchored to the bad format.

## Fix

Add an opt-in `await-listener` property (default `false`, so existing behaviour is unchanged). When enabled, `negotiate()` waits until the downstream caps query changes (a consumer connected and intersected its own caps in) before fixing caps. The format is then chosen **jointly** with the consumer, so it's one both sides support.

Detection watches for the downstream caps **change** rather than inspecting formats, so it stays fully format-agnostic and keeps working as GPUs add new formats. The wayland compositor runs on its own thread, so the application keeps rendering while negotiation is deferred, and `unlock()` aborts the wait so state changes never block on a consumer that never arrives.

## Testing

Verified on real hardware via a producer/consumer interpipe harness (producer started first, consumer attached ~1.5s later, matching Wolf's lifecycle):

- **Intel VAAPI** (vapostproc + vah265enc): default-off reproduces the `not-negotiated` failure; `await-listener=true` negotiates jointly and produces H265 (exit 0).
- **AMD** (the originally reported GPU, vapostproc + vah265enc): `await-listener=true` produces H265 (exit 0); the producer correctly does not fixate early.
- Clean shutdown with `await-listener=true` and no consumer ever attaching (no hang).

Consumed by games-on-whales/wolf to fix DMA-BUF zero-copy negotiation (PR #419).

### DIFF
--- a/gst-plugin-wayland-display/src/waylandsrc/imp.rs
+++ b/gst-plugin-wayland-display/src/waylandsrc/imp.rs
@@ -12,8 +12,9 @@ use gst_video::{NavigationEvent, VideoCapsBuilder, VideoFormat, VideoInfo, Video
 use once_cell::sync::Lazy;
 use std::collections::HashMap;
 use std::ops::DerefMut;
-use std::sync::atomic::{AtomicBool, AtomicPtr, Ordering};
-use std::sync::{Arc, LazyLock, Mutex};
+#[cfg(feature = "cuda")]
+use std::sync::atomic::AtomicPtr;
+use std::sync::{Arc, Condvar, LazyLock, Mutex};
 use tracing_subscriber::Registry;
 use tracing_subscriber::layer::SubscriberExt;
 #[cfg(feature = "cuda")]
@@ -34,8 +35,17 @@ pub struct WaylandDisplaySrc {
     settings: Mutex<Settings>,
     command_tx: Sender<Command>,
     command_rx: Mutex<Option<Channel<Command>>>,
-    /// Set by unlock() to abort a pending await-listener wait in negotiate().
-    await_abort: AtomicBool,
+    /// Coordinates the await-listener wait in negotiate(). `abort` is set by
+    /// unlock() to break the wait; the condvar is notified by event() when a
+    /// downstream RECONFIGURE arrives and by unlock() on abort, so negotiate()
+    /// parks on the event instead of busy-polling.
+    await_state: Mutex<AwaitState>,
+    await_cv: Condvar,
+}
+
+#[derive(Default)]
+struct AwaitState {
+    abort: bool,
 }
 
 impl Default for WaylandDisplaySrc {
@@ -46,7 +56,8 @@ impl Default for WaylandDisplaySrc {
             settings: Mutex::new(Settings::default()),
             command_tx,
             command_rx: Mutex::new(Some(command_rx)),
-            await_abort: AtomicBool::new(false),
+            await_state: Mutex::new(AwaitState::default()),
+            await_cv: Condvar::new(),
         }
     }
 }
@@ -652,16 +663,10 @@ impl BaseSrcImpl for WaylandDisplaySrc {
             // inspecting formats, so this stays fully format-agnostic. The
             // wayland compositor runs on its own thread, so the application keeps
             // rendering while we wait.
-            self.await_abort.store(false, Ordering::SeqCst);
+            self.await_state.lock().unwrap().abort = false;
             let baseline = src_pad.peer_query_caps(None);
             tracing::debug!("await-listener: baseline downstream caps {}", baseline);
             loop {
-                if self.await_abort.load(Ordering::SeqCst) {
-                    return Err(gst::loggable_error!(
-                        CAT,
-                        "Negotiation aborted while awaiting a downstream consumer"
-                    ));
-                }
                 let peer = src_pad.peer_query_caps(None);
                 if !peer.is_empty() && peer != baseline {
                     tracing::info!(
@@ -670,7 +675,22 @@ impl BaseSrcImpl for WaylandDisplaySrc {
                     );
                     break;
                 }
-                std::thread::sleep(std::time::Duration::from_millis(100));
+                let guard = self.await_state.lock().unwrap();
+                if guard.abort {
+                    return Err(gst::loggable_error!(
+                        CAT,
+                        "Negotiation aborted while awaiting a downstream consumer"
+                    ));
+                }
+                // Park until event() signals a downstream RECONFIGURE or
+                // unlock() aborts. The timeout is a safety net: RECONFIGURE is
+                // not guaranteed to cross a decoupling boundary (e.g.
+                // interpipe), so re-check the caps query periodically rather
+                // than relying on the event alone.
+                let _ = self
+                    .await_cv
+                    .wait_timeout(guard, std::time::Duration::from_millis(500))
+                    .unwrap();
             }
         }
         self.parent_negotiate()
@@ -679,16 +699,24 @@ impl BaseSrcImpl for WaylandDisplaySrc {
     fn unlock(&self) -> Result<(), gst::ErrorMessage> {
         // Interrupt a pending await-listener wait in negotiate() so state
         // changes (e.g. shutdown) don't block on a consumer that never arrives.
-        self.await_abort.store(true, Ordering::SeqCst);
+        self.await_state.lock().unwrap().abort = true;
+        self.await_cv.notify_all();
         self.parent_unlock()
     }
 
     fn unlock_stop(&self) -> Result<(), gst::ErrorMessage> {
-        self.await_abort.store(false, Ordering::SeqCst);
+        self.await_state.lock().unwrap().abort = false;
         self.parent_unlock_stop()
     }
 
     fn event(&self, event: &Event) -> bool {
+        // A downstream RECONFIGURE is the real trigger for the await-listener
+        // wait in negotiate(): wake it so it re-checks the caps query
+        // immediately instead of waiting out the poll fallback.
+        if let gst::EventView::Reconfigure(_) = event.view() {
+            let _guard = self.await_state.lock().unwrap();
+            self.await_cv.notify_all();
+        }
         if self.handle_event(&event) {
             return true;
         }
--- a/gst-plugin-wayland-display/src/waylandsrc/imp.rs
+++ b/gst-plugin-wayland-display/src/waylandsrc/imp.rs
@@ -12,7 +12,7 @@ use gst_video::{NavigationEvent, VideoCapsBuilder, VideoFormat, VideoInfo, Video
 use once_cell::sync::Lazy;
 use std::collections::HashMap;
 use std::ops::DerefMut;
-use std::sync::atomic::AtomicPtr;
+use std::sync::atomic::{AtomicBool, AtomicPtr, Ordering};
 use std::sync::{Arc, LazyLock, Mutex};
 use tracing_subscriber::Registry;
 use tracing_subscriber::layer::SubscriberExt;
@@ -34,6 +34,8 @@ pub struct WaylandDisplaySrc {
     settings: Mutex<Settings>,
     command_tx: Sender<Command>,
     command_rx: Mutex<Option<Channel<Command>>>,
+    /// Set by unlock() to abort a pending await-listener wait in negotiate().
+    await_abort: AtomicBool,
 }
 
 impl Default for WaylandDisplaySrc {
@@ -44,6 +46,7 @@ impl Default for WaylandDisplaySrc {
             settings: Mutex::new(Settings::default()),
             command_tx,
             command_rx: Mutex::new(Some(command_rx)),
+            await_abort: AtomicBool::new(false),
         }
     }
 }
@@ -53,6 +56,7 @@ pub struct Settings {
     render_node: Option<String>,
     input_devices: Vec<String>,
     disable_intel_workaround: bool,
+    await_listener: bool,
     #[cfg(feature = "cuda")]
     cuda_context: Option<Arc<Mutex<cuda::CUDAContext>>>,
     #[cfg(feature = "cuda")]
@@ -274,6 +278,18 @@ impl ObjectImpl for WaylandDisplaySrc {
                     )
                     .default_value(false)
                     .build(),
+                glib::ParamSpecBoolean::builder("await-listener")
+                    .nick("Await downstream listener")
+                    .blurb(
+                        "Defer caps negotiation until downstream advertises concrete caps (i.e. a \
+                         consumer is connected). This lets the output format be chosen jointly with \
+                         the consumer instead of fixing an arbitrary default before anything is \
+                         connected. Intended for decoupled setups (e.g. behind an interpipesink) \
+                         where the producer starts before the consumer. The wayland compositor runs \
+                         regardless, so the application can render while negotiation is deferred.",
+                    )
+                    .default_value(false)
+                    .build(),
             ]
         });
 
@@ -337,6 +353,10 @@ impl ObjectImpl for WaylandDisplaySrc {
                 settings.disable_intel_workaround =
                     value.get::<bool>().expect("Type checked upstream");
             }
+            "await-listener" => {
+                let mut settings = self.settings.lock().unwrap();
+                settings.await_listener = value.get::<bool>().expect("Type checked upstream");
+            }
             _ => unreachable!(),
         }
     }
@@ -370,6 +390,10 @@ impl ObjectImpl for WaylandDisplaySrc {
             "disable-intel-workaround" => {
                 let settings = self.settings.lock().unwrap();
                 settings.disable_intel_workaround.to_value()
+            }
+            "await-listener" => {
+                let settings = self.settings.lock().unwrap();
+                settings.await_listener.to_value()
             }
             _ => unreachable!(),
         }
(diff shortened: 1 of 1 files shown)


ABeltramo:
Quick question (I can't look properly at the code rn): what's the difference here with the CUDA pipeline?  

On Nvidia we are also settling for an initial format and then, once the downstream pipeline is plugged we re-negotiate properly and switch to the new format on the fly (with the recent changes that allow the resolution to change in flight). Shouldn't the same apply for DMA?

JBailes:
Same idea, I tried it. The difference is what waylanddisplaysrc advertises. On CUDA it only ever offers BGRA/RGBA, so whatever it settles on is always ingestible and cudaconvertscale handles the rest. The only thing that changes when the encoder plugs in is resolution, which is exactly what the in-flight stuff covers.

DMA advertises the whole DRM format list and settles on the first entry, which is AB4H (ABGR16161616F, a 64-bit float) on both AMD and Intel. vapostproc can't import that at all, it's a hard reject, not a convert. So it'd need a pixel-format switch, not a resolution one. That switch doesn't land in flight: by the time the consumer reconfigures, the capsfilter's already pinned to AB4H and answers the downstream query with it, so we intersect vapostproc's caps against AB4H, get nothing, and the producer errors out before it can re-fixate. The stale buffer's already moving. caps-change-mode=delayed didn't save it either.

await-listener is the same trick as CUDA, just moved earlier: don't fixate until the consumer's there, so the first negotiation is already joint and picks something both sides take. Stays format-agnostic.

If you'd rather keep the two paths identical we can just stop advertising the float/packed formats on DMA and keep the live switch, but that's a denylist to babysit as formats get added, so I went with the timing fix.

ABeltramo:
I'll test it out properly, but I'd say we should have `await-listener` as the default then?  

It's working on CUDA mostly by chance IIUC because luckily we set on a format that can be used downstream, but it should really be fully coordinated from both sides. I can't really see a downside of waiting for a consumer unless it breaks `gst-interpipe` because there's effectively no buffer in the pipeline..?

JBailes:
You are right, I was being too conservative. I'm making await-listener default.

gst-plugin-wayland-display/src/waylandsrc/imp.rs
```diff
@@ -608,9 +632,62 @@ impl BaseSrcImpl for WaylandDisplaySrc {
...
+        // changes (e.g. shutdown) don't block on a consumer that never arrives.
+        self.await_abort.store(true, Ordering::SeqCst);
+        self.parent_unlock()
+    }
+
+    fn unlock_stop(&self) -> Result<(), gst::ErrorMessage> {
+        self.await_abort.store(false, Ordering::SeqCst);
+        self.parent_unlock_stop()
+    }
+
```
ABeltramo, line 711: Shouldn't this be done on `stop`?

gst-plugin-wayland-display/src/waylandsrc/imp.rs
```diff
@@ -608,9 +632,62 @@ impl BaseSrcImpl for WaylandDisplaySrc {
...
+                let peer = src_pad.peer_query_caps(None);
+                if !peer.is_empty() && peer != baseline {
+                    tracing::info!(
+                        "Downstream caps changed (consumer connected), negotiating jointly: {}",
+                        peer
+                    );
+                    break;
+                }
+                std::thread::sleep(std::time::Duration::from_millis(100));
+            }
```
ABeltramo, line 694: Right, so we are stopping the whole Gstreamer pipeline here with a polling loop. Is there really no better alternative?

ABeltramo (commented):
I have to refresh my memory on the caps exchange process in Gstreamer but I feel that this should be available with the `DMA` buffers: we should be able to say we support all these formats, pick one up to create our buffer and then, later on, when the pipeline connects do the negotiation again. Maybe there's a way to force that from the outside or it should be something that we implement in `gst-interpipe`.  

I'd keep this implementation as a last resort option, forcefully blocking the Gstreamer thread doesn't sound like a great solution IMHO..

ABeltramo:
The official docs about the [DMA Buffers](https://gstreamer.freedesktop.org/documentation/additional/design/dmabuf.html?gi-language=c#renegotiation) mention the `renegotiation` mechanism and I think it might be the right path; are we missing support for that in `gst-interpipe`?

gst-plugin-wayland-display/src/waylandsrc/imp.rs
JBailes: unlock_stop() is the right place, it's the designed inverse of unlock(). stop() won't work: on PAUSED->READY the base class flushes (calling unlock()) and joins the streaming task before stop() is called, but that thread is the one parked in the wait, so we'd deadlock waiting to join it. unlock() is exactly the hook for breaking blocking calls. The pair also fires on FLUSH_START/FLUSH_STOP without a state change, so resetting in stop() would leave the flag stuck true and make the next negotiate() abort spuriously after a seek. (negotiate() also re-arms the flag at entry, so the state is always well-defined.)

JBailes:
> The official docs about the [DMA Buffers](https://gstreamer.freedesktop.org/documentation/additional/design/dmabuf.html?gi-language=c#renegotiation) mention the `renegotiation` mechanism and I think it might be the right path; are we missing support for that in `gst-interpipe`?

The renegotiation mechanism is there in gst-interpipe, interpipesrc forwards upstream events to the node and interpipesink's receive_event pushes them (RECONFIGURE included) into the producer pipeline, and interpipesink's get_caps intersects the listeners' caps back to the producer. It's not a missing-feature problem.

The problem is timing. With no listener connected, interpipesink.get_caps advertises "accept anything," so waylanddisplaysrc fixates on AB4H (first DRM format) right away. By the time a consumer attaches and a RECONFIGURE comes back, the capsfilter is pinned and the buffer is already in flight, a pixel-format switch can't land live (this is what I tested; caps-change-mode=delayed didn't help either). A couple of interpipe limits make leaning on renegotiation worse: it reads each listener's already-negotiated caps rather than doing a live query, and event/caps propagation only works with exactly one listener.

await-listener isn't an alternative to renegotiation, it's what lets interpipe's caps broker do its job, by deferring the producer's first fixation until a consumer is actually present so the very first negotiation goes through the listener intersection instead of the "no listeners -> accept anything" path.

gst-plugin-wayland-display/src/waylandsrc/imp.rs
JBailes: It's not that there isn't a better alternative, it's that this is the simpler alternative to a fairly complicated event() solution. A fully non-blocking "defer and return" from negotiate() isn't really on the table: BaseSrc's contract requires negotiate() to either fixate or fail, there's no "not yet, retry me later" return, so refusing to fixate means either erroring out the source or spinning. Blocking the streaming task is the idiomatic way for a live source to say "I have nothing to produce yet." So the realistic improvement is how we wait (event-driven + timeout fallback), not whether we wait.

I'm going to play around with some ideas, but the problem is in the base contract.

ABeltramo:
> With no listener connected, interpipesink.get_caps advertises "accept anything," so waylanddisplaysrc fixates on AB4H (first DRM format) right away.

Right, and that's why we pass the right caps fixed in Wolf at the moment to get around this. I'm not sure this is a better solution, it sounds like it could backfire. Let's test out different ideas and approaches before settling on this as the only solution. 

For example, we haven't properly explored moving `vapostproc` in the upstream pipeline so that they settle on a format and then we pass the va context + `VABuffer` to the downstream pipeline.

Let's explore this for a bit longer..

gst-plugin-wayland-display/src/waylandsrc/imp.rs
JBailes: Okay, rewrote it so that it's a bit better. We're still stuck on some kind of wait, it's just a question of how the wait is implemented. Remember that this is on each individual thread for each individual compositor however. It's not a global wait.

JBailes:
Yeah, that'd move the negotiation into one pipeline where it actually works, and vapostproc would never hand us AB4H, so the format problem goes away. For a single VA consumer it's cleaner than what I've got.

Two things stop me from reaching for it though. It pins format and resolution at the producer, so every consumer gets the same output and a resolution change means renegotiating the shared producer instead of just one branch, which is the fan-out interpipe is there to give us. And passing the VA surface across the boundary isn't free: it's bound to a GstVaDisplay that doesn't propagate across the two pipelines, and the surface pool can recycle out from under the consumer since interpipe lets the pipelines run in different states. A DMABuf fd is much easier to hand over than a context-bound surface.

Worth a look, and the thing to prove first is whether a VA surface even survives the crossing with a shared display and a sane pool.

ABeltramo:
Logging this here just for context, I'm formulating an idea, stay tuned 😜 

The trick on Nvidia is that we are setting the caps to `BGRA` 
https://github.com/games-on-whales/gst-wayland-display/blob/b15285a2f1bb4dae5725b049915a4971664fafc6/gst-plugin-wayland-display/src/waylandsrc/imp.rs#L434-L444

And then downstream `cudaconvertscale` is doing the heavy lifting of turning those into `NV12`

```
2026-06-18T10:27:13.285937Z  INFO gstwaylanddisplaysrc::waylandsrc::imp: Handling context query with CUDA
0:00:17.561051620      10687      10841 DEBUG           interpipesrc gstinterpipesrc.c:484:gst_inter_pipe_src_query:<interpipesrc_5733117292241342039_video> Answered context query across the interpipe boundary
0:00:17.561283240      10687      10841 INFO           interpipesink gstinterpipesink.c:974:gst_inter_pipe_sink_add_listener:<5733117292241342039_video> Adding new listener interpipesrc_5733117292241342039_video
0:00:17.561416450      10687      10841 DEBUG       cudaconvertscale gstcudaconvertscale.c:324:gst_cuda_base_convert_transform_caps:<cudaconvertscale2> transformed video/x-raw(memory:CUDAMemory), format=(string)NV12, width=(int)1920, height=(int)1080, pixel-aspect-ratio=(fraction)1/1, framerate=(fraction)[ 0/1, 2147483647/1 ] into video/x-raw(memory:CUDAMemory), width=(int)[ 1, 2147483647 ], height=(int)[ 1, 2147483647 ], pixel-aspect-ratio=(fraction)[ 1/2147483647, 2147483647/1 ], framerate=(fraction)[ 0/1, 2147483647/1 ]
0:00:17.561505200      10687      10841 INFO           interpipesink gstinterpipesink.c:1014:gst_inter_pipe_sink_add_listener:<5733117292241342039_video> Setting listener caps to video/x-raw(memory:CUDAMemory), format=(string)BGRA, width=(int)1920, height=(int)1080, framerate=(fraction)60/1
0:00:17.561518550      10687      10841 INFO            interpipesrc gstinterpipesrc.c:400:gst_inter_pipe_src_start:<interpipesrc_5733117292241342039_video> Listening to node 5733117292241342039_video
0:00:17.561786651      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:324:gst_cuda_base_convert_transform_caps:<cudaconvertscale2> transformed video/x-raw(memory:CUDAMemory), format=(string)BGRA, width=(int)1920, height=(int)1080, framerate=(fraction)60/1 into video/x-raw(memory:CUDAMemory), format=(string){ I420, YV12, NV12, NV21, P010_10LE, P012_LE, P016_LE, I420_10LE, I420_12LE, Y444, Y444_10LE, Y444_12LE, Y444_16LE, BGRA, RGBA, RGBx, BGRx, ARGB, ABGR, RGB, BGR, BGR10A2_LE, RGB10A2_LE, Y42B, I422_10LE, I422_12LE, RGBP, BGRP, GBR, GBRA, GBR_10LE, GBR_12LE, GBR_16LE, VUYA }, width=(int)[ 1, 2147483647 ], height=(int)[ 1, 2147483647 ], framerate=(fraction)60/1
0:00:17.561834851      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:324:gst_cuda_base_convert_transform_caps:<cudaconvertscale2> transformed video/x-raw(memory:CUDAMemory), format=(string)BGRA, width=(int)1920, height=(int)1080, framerate=(fraction)60/1 into video/x-raw(memory:CUDAMemory), format=(string){ I420, YV12, NV12, NV21, P010_10LE, P012_LE, P016_LE, I420_10LE, I420_12LE, Y444, Y444_10LE, Y444_12LE, Y444_16LE, BGRA, RGBA, RGBx, BGRx, ARGB, ABGR, RGB, BGR, BGR10A2_LE, RGB10A2_LE, Y42B, I422_10LE, I422_12LE, RGBP, BGRP, GBR, GBRA, GBR_10LE, GBR_12LE, GBR_16LE, VUYA }, width=(int)[ 1, 2147483647 ], height=(int)[ 1, 2147483647 ], framerate=(fraction)60/1
0:00:17.561855731      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:324:gst_cuda_base_convert_transform_caps:<cudaconvertscale2> transformed video/x-raw(memory:CUDAMemory), format=(string)BGRA, width=(int)1920, height=(int)1080, framerate=(fraction)60/1 into video/x-raw(memory:CUDAMemory), width=(int)[ 1, 2147483647 ], height=(int)[ 1, 2147483647 ], framerate=(fraction)60/1
0:00:17.561997391      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:1127:gst_cuda_base_convert_fixate_caps:<cudaconvertscale2> trying to fixate othercaps video/x-raw(memory:CUDAMemory), width=(int)1920, height=(int)1080, framerate=(fraction)60/1, format=(string)NV12, pixel-aspect-ratio=(fraction)1/1 based on caps video/x-raw(memory:CUDAMemory), format=(string)BGRA, width=(int)1920, height=(int)1080, framerate=(fraction)60/1
0:00:17.562009211      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:470:gst_cuda_base_convert_fixate_format:<cudaconvertscale2> source format BGRA
0:00:17.562014811      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:480:gst_cuda_base_convert_fixate_format: iterate 1 structures
0:00:17.562021141      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:443:score_value:<cudaconvertscale2> score BGRA -> NV12 = 62
0:00:17.562026241      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:448:score_value:<cudaconvertscale2> found new best 62
0:00:17.562035271      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:756:gst_cuda_base_convert_fixate_size:<cudaconvertscale2> dimensions already set to 1920x1080, not fixating
0:00:17.562042761      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:1163:gst_cuda_base_convert_fixate_caps:<cudaconvertscale2> fixated othercaps to video/x-raw(memory:CUDAMemory), width=(int)1920, height=(int)1080, framerate=(fraction)60/1, format=(string)NV12, pixel-aspect-ratio=(fraction)1/1
0:00:17.587712375      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:1453:gst_cuda_base_convert_set_info:<cudaconvertscale2> BGRA from=1920x1080 (par=1/1 dar=16/9), size 8294400 -> NV12 to=1920x1080 (par=1/1 dar=16/9 borders=0:0), size 3110400
0:00:17.612146798      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:1259:gst_cuda_base_convert_decide_allocation:<cudaconvertscale2> Downstream crop meta support: 0
0:00:17.612159638      10687      11107 DEBUG       cudaconvertscale gstcudaconvertscale.c:1296:gst_cuda_base_convert_decide_allocation:<cudaconvertscale2> Downstream provided CUDA stream
```
(note how this is passing the CUDA context with the new `gst-interpipe`)

JBailes:
Spent some time on the "advertise a fixed format + convert downstream" idea. It works on CUDA for a reason that doesn't carry to VA.

On Nvidia the convert stays inside CUDAMemory (cudaconvertscale BGRA->NV12), so the consumer's first element takes the producer's caps as-is. On VA the downstream convert is DMABuf→VAMemory, a memory-feature change. I pinned the producer to AR24 (no AB4H), put vapostproc in the consumer, attached a listener late: not-negotiated. The AR24 caps cross interpipe fine, but interpipesrc replays the producer's already-fixed caps and starts pushing buffers without a live query, so vapostproc never gets to negotiate the import. Same with any format, it's the late attach, not the format.

await-listener fixes exactly that. Defer fixation, the consumer's vapostproc is present for the first negotiation, producer settles on AR30 (not AB4H), vapostproc converts to VAMemory NV12, encodes clean. Confirmed against the real waylanddisplaysrc (waylandsink client feeding it) on Intel UHD 770 and AMD/radeonsi.

So curating the advertised formats is worth doing as hardening, stop offering the float formats, but on VA it's not a standalone fix, it still needs the consumer there at fixation.

On fan-out: a second listener breaks regardless of where the convert lives:
Could not send event upstream, more than one listener is connected followed by a vaEndPicture crash. interpipe only propagates events for one listener, so multi-consumer renegotiation isn't on the table today either way. One listener per virtual display works, which is what we actually run.

ABeltramo:
> I pinned the producer to AR24 (no AB4H), put vapostproc in the consumer, attached a listener late: not-negotiated.

The point of that solution is that you would pin the producer to the right format that would then be accepted downstream by `vapostproc`. Obviously if you pin to the wrong format it's not going to work unless I'm missing something..

JBailes:
AR24 isn't the wrong format. vapostproc lists it in its sink caps (AR24:0x0100000000000002, next to AR30/AB24/XB24), it ingests it fine. The not-negotiated wasn't vapostproc rejecting the format.

The producer's a GL renderer, its whole DMA list is packed RGB: AB4H, XB4H, AR30, XR30, AB30, AR24, AB24, XR24... no NV12, no YUV anywhere. So whatever you pin, you still need vapostproc to reach NV12 for the encoder. There's no format you can pin that skips the convert.

What fails is the late attach. interpipesrc takes the producer's already-fixed caps and starts pushing buffers before vapostproc negotiates its VAMemory output, so the convert never gets set up. Pick any of those RGB formats, same result.

It's the timing, not the format: flip await-listener on, pin nothing. Producer defers, the consumer's vapostproc is there for the first negotiation, it settles on AR30 (same RGB family, also "accepted by vapostproc") and encodes clean. Only thing that changed is the consumer was present when it fixated.

Also, and I want to be clear about this: Much of what you are proposing is *complementary* to await, not *contradictory*. Await fixes the timing issue we have with fixate and negotiation, it doesn't fix anything else

ABeltramo:
> Producer defers, the consumer's vapostproc is there for the first negotiation, it settles on AR30 (same RGB family, also "accepted by vapostproc") and encodes clean.

So what happens if you pin `AR30` instead of `AR24` without `await-listener` on?

JBailes:
Just did a quick test where I pinned AR30. It had the same failure as AR24, pin AR30, no await, consumer attaches late, 390 buffers out of the producer, consumer dies with not-negotiated, zero encoded. Identical to AR24.

I think we need to understand the fundamental issue this PR tries to fix better: There is a very limited window for negotiation normally. This is not something that is affected by pinning anything, or other issues you have brought up. What you have brought up is *extremely valid* and, at least IMO, needs to happen. But it's not addressing the fundamental issue this PR tries to resolve, which is extending the window for negotiation.

The underlying problem is with the contract, and admittedly, this is a bandaid over the contract.

JBailes:
Lots of conversation over discord. TL;DR: We're trying to fix an issue with Gstreamer here.

PR #36, opened 2026-06-19
fix(cuda): take own ref on GstCudaContext to fix double-unref at teardown

salty2011:
## Problem

`gst_cuda_ensure_element_context()` and `gst_cuda_handle_set_context()` both open with:

```c
/* If we had context already, will not replace it */
if (*cuda_ctx)
  return TRUE;
```

They report success while transferring nothing. Every other success path is `(transfer full)` into the slot. `CUDAContext::drop` unconditionally unrefs `self.ptr`, so a wrapper built off that short-circuit destroys a reference it never owned.

The compositor hits this on its own context:

- `new_from_gstreamer()` calls the helper, which writes the slot then publishes the context.
- That publish re-enters `waylanddisplaysrc::set_context()` **synchronously**, from inside the same call.
- At re-entry the settings slot is still `None`, so the **inner, non-owning** wrapper is stored and the **outer** one, holding the single legitimate reference, is dropped when the outer call returns.
- The session then runs with the context held only by the published `GstContext`. At teardown that drops, the refcount hits zero, and the `GstCudaContext` is finalized while the pipeline's stored `GstContext` still points at it.
- Disposing the pipeline walks `element->contexts` and lands on freed memory: SIGSEGV in `g_type_check_instance_is_fundamentally_a`, at the end of every session teardown.

Only bites when nothing else outlives the pipeline. A host that pins one `GstCudaContext` for the process lifetime keeps the refcount above zero and the defect stays invisible.

## Fix

`adopt_slot_ref()` takes a reference **only** on the short-circuit, at both constructors.

Detection is unambiguous because both helpers write the slot only when it was NULL on entry: success with a non-NULL pre-value is necessarily the short-circuit, success with a NULL pre-value is necessarily a transfer. The `(transfer full)` paths are untouched.

## What changed since this PR was opened

- **Retargeted from `feat/nv12-output` to `stable`.** That branch is divergent and unmerged; the defect is still present on `stable` at both constructors.
- **The fix approach changed.** The original was an unconditional `gst_object_ref()` before `gst_cuda_stream_new()`. That fixes the SIGSEGV but adds a reference nothing removes on the normal transfer path, turning the crash into a per-session `GstCudaContext` leak. The conditional version is correct in both paths and is what has been running downstream.

## Not addressed here

The slot itself is never cleared when its context dies, so a later call can read a stale pre-value. That hazard predates this change and is unaffected by it. Happy to take it separately if you want it closed.

## Verified

`cargo fmt` clean, `cargo check --features cuda` clean, 23 tests pass. `CUDAContext::new()`'s own `gst_cuda_stream_new` is unaffected (no slot, no short-circuit possible) and is untouched.

### DIFF
--- a/wayland-display-core/src/utils/allocator/cuda/mod.rs
+++ b/wayland-display-core/src/utils/allocator/cuda/mod.rs
@@ -386,11 +386,75 @@ impl CUDAContext {
         })
     }
 
+    /// Make the slot's `GstCudaContext` safe for a [`CUDAContext`] wrapper to OWN.
+    ///
+    /// Both `gst_cuda_ensure_element_context()` and `gst_cuda_handle_set_context()` open
+    /// with the same short-circuit (gst-plugins-bad `gstcudautils.c`):
+    ///
+    /// ```c
+    ///   /* If we had context already, will not replace it */
+    ///   if (*cuda_ctx)
+    ///     return TRUE;
+    /// ```
+    ///
+    /// i.e. they report SUCCESS while transferring NOTHING. Every other SUCCESS path is
+    /// `(transfer full)` into the slot. [`CUDAContext::drop`] unconditionally
+    /// `gst_object_unref()`s `self.ptr`, so a wrapper built off that short-circuit
+    /// destroys a reference it never owned.
+    ///
+    /// That is exactly what happens on the compositor's OWN context.
+    /// `new_from_gstreamer()` calls `gst_cuda_ensure_element_context()`, which creates the
+    /// context (rc=1, into the slot) and publishes it -- and that publish re-enters
+    /// `waylanddisplaysrc::set_context()` SYNCHRONOUSLY, from inside the very call, with the
+    /// context it is still in the middle of creating. Confirmed live by a leaks-tracer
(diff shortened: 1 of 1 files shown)


JBailes:
lgtm!

salty2011:
May want to test it, I would but have several projects currently monopolising the hardware 😂

ABeltramo:
Can you drop all the lint edits and just keep the fix itself so that we don't have merge conflicts?

salty2011:
Sure no problem, jumping on it

salty2011:
@ABeltramo done, should now only have the required fix

salty2011:
Retargeted from `feat/nv12-output` (divergent, unmerged) to `stable`, where the defect is still present. Also swapped the fix: the original unconditional `gst_object_ref` traded the SIGSEGV for a per-session context leak on the normal transfer path. The conditional version is correct in both paths. Rationale and refcount walk in the body.

ABeltramo:
Sorry guys this has taken me this long to review, I'm not a fan of the ~50 lines of comment, but it's actually quite useful in this case 😅 
LGTM!

PR #37, opened 2026-06-20
feat(nv12): Vulkan NV12 + VA/Vulkan encode

JBailes:
Vulkan RGBA to NV12 converter. waylanddisplaysrc feeds VA encoders an NV12 dmabuf direct, no vapostproc.

GLES renders RGBA. Vulkan imports that dmabuf (VK_EXT_external_memory_dma_buf), compute-shades BT.601 NV12, exports an NV12 dmabuf the encoder takes zero-copy. va_share binds it to the encoder's gst-va display so the encoder reuses one surface per ring slot instead of importing a fresh one each frame.

Pipeline:

```
waylanddisplaysrc ! vah265enc
```

A VA encoder directly downstream asks for NV12, so it negotiates on its own. Behind a format-agnostic sink like interpipesink (Wolf), set `nv12=true` so the source offers NV12 first. RGBA stays as a fallback.

Tested to EOS, 20 frames:
- AMD 7900 XTX, vah265enc
- Intel UHD 770, vah265lpenc
- Nvidia 5080, converter exports NV12

Real content on Intel (SMPTE + moving ball, decoded): clean, no garbling or tearing.

Default build. CUDA path stacks on top.

## Vulkan encode

Second consumer off the same converter: vulkanh264enc, zero-copy.

waylanddisplaysrc emits NV12 `memory:VulkanImage` on the encoder's shared GstVulkanDevice (via gst context). The encoder views the image, no copy, no dmabuf hop. Set `vulkan=true`.

Pipeline (Wolf, one process):

```
waylanddisplaysrc vulkan=true ! interpipesink
interpipesrc ! vulkanh264enc
```

Fan-out works: one capture, N encoders, mixed profiles. Seed the image's tracked layout to VIDEO_ENCODE_SRC and null its per-memory timeline semaphore so each encoder reads dependency-free; gate ring reuse on buffer release.

Tested on Nvidia 5080, gst 1.28.4 and 1.29.1:
- one consumer, ~170 frames
- two consumers, high + main, ~206 each

Needs the vkh264enc DPB patch (create_dpb_pool in new_sequence) and Vulkan headers >= 1.4.317 at build.

PR #38, opened 2026-06-21
ci: build/test/benchmark harness + CI workflow (extracted from #37)

JBailes:
## What

Extracts the **build/test/benchmark harness and its CI workflow** out of the Vulkan NV12 encode PR (#37) so the reusable test/CI infrastructure can be reviewed and merged on its own, shrinking #37 to just the feature.

- **`ci/harness.sh`** — one phased entry point (`detect · lint · build · unit · gpu · integration · bench`) that detects the GPU(s) present, builds the workspace (default + `cuda` when an Nvidia GPU is present), runs the tests, and prints a single pass/fail summary. It encodes the bare-box environment knowledge: `LIBCLANG_PATH` for bindgen, `XDG_RUNTIME_DIR` for the compositor, render-node→vendor detection, per-vendor encoder elements.
- **`ci/README.md`** — harness docs (phases, vendor matrix, host prerequisites).
- **`.github/workflows/harness.yml`** — runs `ci/harness.sh lint build unit` on a GitHub-hosted runner (in the `ghcr.io/games-on-whales/gstreamer` container); the full `gpu/integration/bench` matrix runs on a self-hosted `gpu` runner, gated by `vars.GPU_RUNNER == 'true'`.

## Why separate

The harness is reusable infrastructure that's valuable independent of the Vulkan feature: on a CPU-only runner it exercises **lint + build + the existing unit tests** (the repo had no build/unit CI before — only `rust-format.yml`). The GPU/bench phases **self-skip** when there's no render node, so this is safe to land first.

## Scope (what's intentionally NOT here)

The feature-specific pieces the `gpu`/`bench` phases drive — `gst-plugin-wayland-display/tests/encode.rs` and `benchmark/` — call into the Vulkan NV12 code (`waylanddisplaycore::utils::vulkan_nv12::supported_nv12_modifiers`, etc.) and would not compile on `stable`, so they **stay in #37**. The harness self-skips them until they exist. #37 can rebase onto this and drop these three files (they're byte-identical to its versions, so the rebase is clean).

## Verification

This PR's own `harness.yml` run is the proof: `ci/harness.sh lint build unit` against `stable` in the gstreamer container. The `harness-gpu` job stays inert (no `GPU_RUNNER`).

Follow-up (not in this PR): `rust-format.yml` is now subsumed by the harness `lint` phase and could be removed.

### DIFF
--- a/.github/workflows/harness.yml
+++ b/.github/workflows/harness.yml
@@ -1,0 +1,50 @@
+name: build & test
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  # CPU phases on a GitHub-hosted runner: lint, build (default + the cuda build
+  # is skipped without an Nvidia GPU), and the non-ignored tests. The GPU phases
+  # (gpu / integration / bench) self-skip when there is no render node.
+  harness:
+    name: harness (lint · build · unit)
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/games-on-whales/gstreamer:1.26.7
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: System dependencies
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          apt-get update
+          apt-get install -y --no-install-recommends \
+            build-essential pkg-config curl ca-certificates git libssl-dev libicu76 \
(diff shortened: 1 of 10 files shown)


JBailes:
@ABeltramo Extracted the test harness and CI work out of 37.

PR #44, opened 2026-07-26
fix(compositor): pointer focus for newly mapped windows — bbox refresh, map-time enter, and a get_pointer-race guard

salty2011:
`Window::bbox()` stays `(0,0)` until `on_commit()` recomputes it from the surface tree, and the per-commit `on_commit()` at the top of `commit()` only runs for surfaces **already in the space**. A client whose root toplevel is mapped by this commit and never re-committed afterwards therefore keeps an empty bbox forever, so `Space::element_under()` never resolves it and `wl_pointer` focus is never assigned. Keyboard focus is assigned directly at map, which is what makes this awkward to diagnose: typing works, the pointer is dead.

This PR originally carried only the bbox refresh, and its description said of the map-time-enter follow-up: *"That is a protocol-visible behaviour change rather than a bug fix, so it seemed wrong to bundle it in."* We have since shipped that follow-up in production, hit the failure mode it introduces, and root-caused it — so the PR now carries the complete, hardened change as two further commits, each independently green including the test updates it necessitates. Field evidence below.

### The three commits

1. **`on_commit()` after `map_element()`** (unchanged from the original PR): recomputes the bbox from the mapping commit's buffer so `element_under()` can resolve the window. The in-tree `fixture.rs` workaround ("I've spent hours trying to debug why the window size was kept at (0,0)") is this same call done by hand.

2. **Synthetic zero-delta `pointer_motion()` at map**: delivers `wl_pointer.enter` to the newly mapped toplevel immediately and runs constraint activation, so a pointer lock/confine requested *before* the surface became mappable (nested gamescope with `--force-grab-cursor`) is applied the moment the surface is focusable. Pointer focus thereby follows the newest toplevel exactly like keyboard focus. `move_mouse` is updated in the same commit for the earlier enter — the behaviour change the original description predicted — by draining the map-time `Enter`; its motion-delivery assertions are unchanged.

3. **An edge-triggered refocus guarding the race that (2) introduces.** smithay emits `enter` only on a focus *transition*, broadcasts it to the `wl_pointer` resources existing at that instant, and records the new focus unconditionally (`GetPointer` sends no retroactive enter). So when (2) resolves focus at map time, a client that calls `wl_seat.get_pointer` *after* that instant loses permanently: the enter reached zero resources, and every subsequent motion takes smithay's same-target arm — `wl_pointer.motion` only, no enter, forever. The client never learns it has pointer focus and discards all pointer input.

   **This is not theoretical.** Rootful Xwayland (`Xwayland :0`, one surface for the whole X screen, mapped exactly once) loses the race deterministically: the cursor moves (compositor-drawn) but every X `ButtonPress` lands at one frozen stale coordinate — "mouse moves, nothing responds to clicks" on an XFCE-in-Xwayland desktop. gamescope loses it intermittently and self-heals on its next surface re-map, which presents as "input dead at launch, starts working later".

   The fix: the map handler arms `pending_pointer_refocus` *after* its synthetic motion (order matters — that motion is itself a `pointer_motion()` call and would consume the flag inside the race window). The next real motion, guarded on surface-present (else retained), pointer-not-grabbed (else retained — a forced leave mid-click-drag would break the drag), and no-active-constraint (else cleared — a constraint can only activate on an entered surface, so the client demonstrably has its enter), forces the focus to `None` so the following motion takes the `(focus, None)` arm and re-emits a genuine `enter`. The re-emitted enter's serial is allocated after the forced leave's so it is the newer one, since clients echo enter serials (`set_cursor`, constraint activation).

   Two tests assert a real fixture client receives the re-emitted `Enter`, and that the flag survives a motion with nothing under the pointer. (The fixture client binds its pointer at seat-bind time so it cannot lose the race naturally; the tests arm the flag directly.)

### Field verification (Quasar, production fork)

Debugging chain on a live deployment, each hop instrumented: evdev capture proved motion+buttons reached the compositor's input devices; `WAYLAND_DEBUG=1` inside the client container showed `wl_pointer.motion` and `wl_pointer.button` delivered with **zero `wl_pointer.enter`** in the whole trace (while `wl_keyboard.enter` arrived fine); `xinput test-xi2 --root` inside Xwayland showed `Motion=0, RawMotion=0`, ButtonPress frozen at one coordinate. With (3) applied: X `Motion` events flow and ButtonPress tracks the cursor; Steam/gamescope validated over repeated cold starts with no dead-input window; both confirmed by a human tester.

The deeper root cause — no retroactive `enter` for a `wl_pointer` created while its surface already holds focus — arguably belongs in smithay's `GetPointer` handling; we intend to raise it there. This PR's (3) is the compositor-side guard that makes map-time focus safe regardless.

### Verification

`cargo test -p wayland-display-core --lib` in a container with the build deps, **at every commit** (the series is bisect-safe — each commit carries the test updates it necessitates):

| | result |
|---|---|
| base (`stable` + the bbox commit) | 18 passed, 1 failed (`utils::allocator::tests::test_dmabuf` — needs a real render node, fails in a deviceless container regardless) |
| + map-time enter | 18 passed, 1 failed — `move_mouse` green under the new behaviour |
| + refocus guard (tip) | **21 passed**, 1 failed — the three refocus tests green (`pending_refocus_re_emits_enter` asserts a real fixture client receives the re-emitted `Enter`; `..._retained_without_surface` and `..._retained_while_grabbed` cover the retain guards) |

`cargo fmt --all --check` clean and clippy zero-new at both commits.

### DIFF
--- a/wayland-display-core/src/comp/input.rs
+++ b/wayland-display-core/src/comp/input.rs
@@ -199,12 +199,66 @@ impl State {
         delta_unaccelerated: Point<f64, Logical>,
     ) {
         self.last_pointer_movement = Instant::now();
-        let serial = SERIAL_COUNTER.next_serial();
         let pointer = self.seat.get_pointer().unwrap();
-        let under = self
+        let under: Option<(FocusTarget, Point<f64, Logical>)> = self
             .space
             .element_under(self.pointer_location)
             .map(|(w, pos)| (w.clone().into(), pos.to_f64()));
+
+        // Edge-triggered pointer refocus.
+        //
+        // smithay only emits wl_pointer.enter on a focus TRANSITION, and broadcasts it to
+        // the wl_pointer resources that exist at that instant while recording the new focus
+        // unconditionally. This compositor resolves pointer focus once, via a synthetic
+        // zero-delta motion in the toplevel map handler. A client that calls
+        // wl_seat.get_pointer AFTER that instant loses the race: the enter reached zero
+        // resources, smithay's focus is already Some(surface), and every later motion takes
+        // smithay's same-target arm (wl_pointer.motion only) -- so the client never learns
+        // it has pointer focus and ignores all pointer input, forever. Rootful Xwayland
+        // (maps exactly once) fails deterministically; gamescope fails intermittently.
+        //
+        // Fix: on the first real motion after a map, force smithay's focus to None so the
+        // motion below takes the (focus, None) => enter arm and re-emits the enter.
+        //
+        // The flag is RETAINED (not consumed) when there is no surface under the pointer --
+        // there is nothing to refocus onto yet -- and while the pointer is grabbed, since
+        // smithay's default click grab is active for as long as a button is held and a
+        // forced leave mid-click-drag would break the drag.
+        let refocus_target = under
+            .as_ref()
+            .filter(|_| self.pending_pointer_refocus && !pointer.is_grabbed());
+        if let Some((target, _)) = refocus_target {
+            // An active pointer constraint proves the client already received its enter
+            // (constraints only activate on an entered surface), so the refocus is
+            // unnecessary rather than deferred -- clear the flag, don't retry.
+            let constrained = target
+                .wl_surface()
+                .map(|surface| {
+                    with_pointer_constraint(
+                        &surface,
+                        &pointer,
+                        |constraint| matches!(constraint, Some(c) if c.is_active()),
+                    )
+                })
+                .unwrap_or(false);
+
+            self.pending_pointer_refocus = false;
+            if !constrained {
+                pointer.motion(
+                    self,
+                    None,
+                    &MotionEvent {
+                        location: self.pointer_location,
+                        serial: SERIAL_COUNTER.next_serial(),
+                        time: event_time_msec,
+                    },
+                );
+            }
+        }
+
+        // Allocated AFTER the refocus block: the re-emitted wl_pointer.enter below must carry
+        // a serial NEWER than the forced leave's, since clients echo enter serials back.
+        let serial = SERIAL_COUNTER.next_serial();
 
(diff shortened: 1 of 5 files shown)


salty2011:
hold off on merge with this.

Testing the GOW XFCE image and found an issue that brakes input capture from correctly working. however this does seem to make gamescope more reliable.

salty2011:
Amended: this now carries the follow-up the original description deliberately left out (the map-time enter), plus a guard for a race it introduces that we hit in production — a client calling `wl_seat.get_pointer` after the map-time focus resolution never receives its `enter` and discards all pointer input (rootful Xwayland deterministically; nested gamescope intermittently). Full story and field evidence in the updated description. Each commit is independently green, including the `move_mouse` expectation change the original text predicted.

JBailes:
@ABeltramo This is a merge to stable, passing it to you. My thoughts.

Make sure this is rebased against stable
Delete the fixture workaround
We need a constraint guard test

salty2011:
Rebased onto `stable`. Both points done.

* **Fixture workaround gone.** New `WaylandClient::recreate_pointer()` releases and re-requests the client's `wl_pointer`, so the tests drive the real late-pointer race. No test touches server state.
* **Constraint guard test added.** Asserts an active lock suppresses the forced leave and *clears* the trigger rather than deferring it.
* **Dropped one test.** The no-surface case isn't constructible through the public API: windows map at (0,0) where the pointer starts, and `under` resolves pre-move, so the first motion after a map always finds a surface. That behaviour is emergent from the `filter`, not a written branch. Can add fixture window placement if you want it covered.

Each commit independently green. 26 passed, fmt clean.

salty2011:
Sorry, just realised this was sitting in Draft, changing to Ready for review.

ABeltramo:
LGTM! Thanks guys!

PR #45, opened 2026-07-26
fix(vulkan): clear PTS on recycled encode-src ring buffers so do_timestamp restamps

salty2011:
`VulkanNv12::to_gst_buffer()` returns a bare clone of the current ring slot's persistent `GstBuffer`. `waylanddisplaysrc` runs with `set_do_timestamp(true)`, which only stamps a buffer whose PTS is `GST_CLOCK_TIME_NONE` — and a recycled slot buffer still carries the PTS from its previous use, so `do_timestamp` skips it.

With the default ring depth of 4, the exported PTS therefore degenerate to four recurring values, non-monotonic and duplicated across frames. Live trace on my host: **99.9% of frames shared a timestamp with another frame, 25% were non-monotonic, and exactly four distinct RTP timestamps cycled forever.**

Downstream that is fatal for WebRTC: `rtph264pay` derives RTP timestamps from those PTS, and libwebrtc's `PacketBuffer` cannot assemble the result. The browser dropped roughly **60% of frames at the decode stage** on a clean LAN with zero packet loss (`framesAssembledFromMultiplePackets` 39.7%, versus 99.9% on the VA path). The stream looks black or frozen while `video.videoWidth` is set and `currentTime` advances, so it is easy to misread as a negotiation problem.

The RGBx and DMABuf output paths never hit this because each builds a fresh `GstBuffer` (PTS `NONE`) every frame. This restores the same behaviour for the Vulkan path: clear PTS/DTS/duration on the recycled buffer before returning it, so `do_timestamp` re-stamps it with the current running time each frame.

Scoped to the `encode_src` case, so the compositor-NV12 dmabuf path (`encode_src == false`) is byte-identical. Only timing metadata is reset — the buffer's memory and video meta are untouched.

Build-verified on this branch's head (`43d4c25`). This fix has been running in my deployment since 2026-07-06.

### DIFF
--- a/wayland-display-core/src/utils/vulkan_nv12.rs
+++ b/wayland-display-core/src/utils/vulkan_nv12.rs
@@ -1462,8 +1462,31 @@ impl VulkanNv12 {
 
     /// The just-converted NV12 export slot as a gst buffer. Returns the slot's cached
     /// buffer (ref-counted) so the VA encoder reuses one stable surface per slot.
+    ///
+    /// The slot's `GstBuffer` is allocated once and reused every `RING` frames, so a
+    /// bare `.clone()` also carries the timestamps stamped onto it on its *previous* use.
+    /// `waylanddisplaysrc` runs `set_do_timestamp(true)`, which only stamps a buffer whose
+    /// PTS is `NONE`; a recycled buffer whose PTS is already set is left untouched — so the
+    /// exported timestamps degenerate to the ring's ~4 recurring values (non-monotonic,
+    /// duplicated across frames), which a downstream RTP payloader turns into RTP timestamps
+    /// that libwebrtc's PacketBuffer cannot assemble (VULKAN-WORKLOG 2026-07-06 root cause).
+    /// The RGBx/DMABuf output paths never hit this because they build a *fresh* `GstBuffer`
+    /// (PTS `NONE`) every frame. Clear the recycled buffer's timing metadata so BaseSrc's
+    /// `do_timestamp` re-stamps it with the current running-time each frame, exactly like
+    /// those paths. Only touches PTS/DTS/duration — the memory and video meta are untouched.
+    ///
+    /// Scoped to the `encode_src` (Vulkan `memory:VulkanImage` → `vulkanh264enc`) path only.
+    /// The VA compositor-NV12 dmabuf path (`encode_src == false`, GW-02) is left byte-identical
+    /// as it rides clean today; only the vulkan output path's timestamping changes.
     pub fn to_gst_buffer(&self) -> Result<GstBuffer, Err> {
-        Ok(self.outputs[self.cur].buffer.clone())
+        let mut buffer = self.outputs[self.cur].buffer.clone();
+        if self.encode_src {
+            let b = buffer.make_mut();
+            b.set_pts(gst::ClockTime::NONE);
+            b.set_dts(gst::ClockTime::NONE);
+            b.set_duration(gst::ClockTime::NONE);
+        }
+        Ok(buffer)
     }
 }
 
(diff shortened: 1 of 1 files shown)


salty2011:
Housekeeping: this is independent of my other two PRs on this branch (#46 and #47). I verified all six pairwise orderings apply cleanly, and all three applied together build clean on `43d4c25` — so they can be merged in any order, individually, with no rebase needed. I split them rather than sending one PR so the small self-contained fixes are not held up by discussion of the larger one. Happy to combine them into a single PR if you would prefer one thread.

JBailes:
Diagnosis holds, and the four-distinct-timestamps trace against RING = 4 is close to conclusive.

The fix disables the fan-out gate. make_mut() at refcount 2 copies rather than returning the original, so to_gst_buffer hands back a new GstBuffer each frame and outputs[cur].buffer sits at refcount 1 forever. convert_inner gates slot reuse on exactly that:

while self.outputs[idx].buffer.get_mut().is_none() {

That loop is the only thing stopping the producer overwriting a slot's VkImage while a consumer reads it. It now passes on the first check, always, and it fails in the direction that looks fine: the copy refs the same memory, so zero-copy survives and the interlock does not. Confident on the gst semantics, unverified at runtime. Count that loop's iterations before and after and it settles.

Instead, reset in convert_inner right after the gate, where the slot is known uniquely owned:

if let Some(b) = self.outputs[idx].buffer.get_mut() {
    b.set_pts(gst::ClockTime::NONE);
    b.set_dts(gst::ClockTime::NONE);
    b.set_duration(gst::ClockTime::NONE);
}

Against my own suggestion: your copy makes the returned buffer legally writable, and basesrc is currently doing an aliased PTS write on a refcount-2 object. Mine reintroduces that. I still prefer it, because an aliased write nobody reads is benign and an unguarded GPU write hazard is not.

The VA path should have this bug too. Both buffer types call the same to_gst_buffer into the same basesrc and nothing resets PTS on the NV12 side, yet you have it riding clean. My guess is that comparison was RGBx plus vapostproc rather than this converter. Worth settling, because if the mechanism is generic the reset belongs unconditional rather than behind encode_src.

Smaller: the doc's first line still promises the slot's cached buffer, which this path no longer returns, and VULKAN-WORKLOG 2026-07-06 is not in this repo.

Move the reset into convert_inner, and tell me what the VA measurement ran against.

salty2011:
Fix moved as you suggested, and the loop count settles it.

* Reset now lives in `convert_inner` right after the reuse gate. `to_gst_buffer` is back to a bare clone. The `make_mut()` copy and its gate-blinding are gone.
* Measured both builds on an RTX 5090 with a forced-contention pipeline (60 fps producer, ~33 fps consumer via `identity sleep-time=30000`, `queue max-size-buffers=8` > RING=4), counting gate-loop iterations per frame over 900 frames:

| build | frames where the gate waited | total 100us iterations |
|---|---|---|
| PR as submitted (`make_mut`) | 0 / 900 | 0 |
| this rework | 867 / 900 | 202,677 |

Your prediction exactly: the copy blinds the gate under contention. Uncontended steady state is 0 waits on both, as expected.
* One deliberate gap: on the gate's 1s-timeout escape the slot is still shared, so the reset skips (with a warn log) and that frame keeps a stale PTS. The stream is already degraded there and a write to a shared header seemed worse. If you would rather fail the frame than reuse the slot on timeout, happy to do that as a follow-up since it changes the gate's semantics.
* You guessed the VA measurement right. It ran against the production VA path, RGBx into vapostproc, fresh buffer every frame. It never touches this converter. The `encode_src == false` mode of this converter likely shares the mechanism, but it has no reuse gate, so an unconditional reset would silently skip under contention. Kept it scoped.
* Doc first line fixed, worklog reference removed.

PR #46, opened 2026-07-26
fix(vulkan): actually fall back to tiled when a LINEAR encode-src alloc fails

salty2011:
`WOLF_VULKAN_LINEAR_ENCSRC`'s comment promises the knob is safe because it "falls back to tiled if the linear alloc fails", but the code does not do that: when `gst_vulkan_image_memory_alloc_with_image_info` returns null for the LINEAR image, `alloc_encode_src_buffer` simply returns `None`. That fails `VulkanNv12::new_on_shared` ("encode-src image allocation failed") and the session's entire Vulkan output fails to be created.

Verified live on an RTX 5090 (2026-07-24): NVIDIA's Vulkan-Video encoder **rejects** a LINEAR encode-src image, so setting the knob on an NVIDIA host hard-failed every session instead of degrading gracefully.

This implements the documented behaviour: if the LINEAR allocation returns null while `WOLF_VULKAN_LINEAR_ENCSRC` is set, warn and retry the identical allocation with `vk::ImageTiling::OPTIMAL`. A genuine allocation failure still returns `None` after the retry, and the non-knob path is byte-identical because the retry branch is gated on `linear_encsrc`.

Scope of the safety claim: the fallback only covers a driver that rejects the LINEAR image at allocation time. A driver that accepts the LINEAR image and then corrupts or fails at `vkCmdEncodeVideo` is not covered. So on a mixed-GPU fleet the knob degrades gracefully for allocation-time rejection (the NVIDIA case above), nothing more.

Build-verified on this branch's head (`43d4c25`).

### DIFF
--- a/wayland-display-core/src/utils/vulkan_share.rs
+++ b/wayland-display-core/src/utils/vulkan_share.rs
@@ -483,11 +483,26 @@ pub fn alloc_encode_src_buffer(
             .initial_layout(vk::ImageLayout::UNDEFINED)
             .push_next(&mut profile_list);
 
-        let mem_ptr = gstvk::gst_vulkan_image_memory_alloc_with_image_info(
+        let mut mem_ptr = gstvk::gst_vulkan_image_memory_alloc_with_image_info(
             device.to_glib_none().0,
             &image_info as *const vk::ImageCreateInfo as *mut _,
             vk::MemoryPropertyFlags::DEVICE_LOCAL,
         );
+        if mem_ptr.is_null() && linear_encsrc {
+            // The tiled-fallback promised above: some encoders (NVIDIA Vulkan-Video)
+            // reject a LINEAR encode-src image outright, so a failed LINEAR alloc must
+            // not kill the whole vulkan output -- retry with the tiled default.
+            tracing::warn!(
+                "vulkan_share: LINEAR encode-src alloc failed (WOLF_VULKAN_LINEAR_ENCSRC) -- \
+                 falling back to tiled (OPTIMAL)"
+            );
+            let image_info = image_info.tiling(vk::ImageTiling::OPTIMAL);
+            mem_ptr = gstvk::gst_vulkan_image_memory_alloc_with_image_info(
+                device.to_glib_none().0,
+                &image_info as *const vk::ImageCreateInfo as *mut _,
+                vk::MemoryPropertyFlags::DEVICE_LOCAL,
+            );
+        }
         if mem_ptr.is_null() {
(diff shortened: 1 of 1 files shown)


salty2011:
Housekeeping: this is independent of my other two PRs on this branch (#45 and #47). I verified all six pairwise orderings apply cleanly, and all three applied together build clean on `43d4c25` — so they can be merged in any order, individually, with no rebase needed. I split them rather than sending one PR so the small self-contained fixes are not held up by discussion of the larger one. Happy to combine them into a single PR if you would prefer one thread.

JBailes:
Approve, merge it.

I concur on the bug: alloc_encode_src_buffer returned None on a null LINEAR alloc, create_encode_output hit .ok_or("encode-src image allocation failed"), and Vulkan output died. The knob comment promised a tiled fallback that did not exist.

The patch is right. gst_vulkan_image_memory_alloc_with_image_info returns NULL when vkCreateImage fails, ash's .tiling() preserves the pushed VkVideoProfileListInfoKHR chain, the retry is gated on linear_encsrc, and the fallback is the existing default path: LINEAR scratch -> OPTIMAL encode-src. RING=4, so the warning is capped at 4 lines at init.

Two small asks before merge:

Scope the safety wording. "Safe to leave enabled on a mixed-GPU fleet" only covers drivers rejecting LINEAR at allocation time. It does not cover a driver accepting the image then corrupting or failing at vkCmdEncodeVideo, which matters because the GFX12 swizzle bug is silent corruption. Say LINEAR allocation failing in the PR body and code comment.

Move the info! to log the effective tiling after allocation. Right now fallback hosts log tiling=LINEAR four times, then retract it four times.

Follow-up only: probing via failed vkCreateImage is invalid usage. vkGetPhysicalDeviceImageFormatProperties2 with the same video profile chain is the spec path.

No test is possible without a GPU that rejects LINEAR and the GPU CI job is skipped, so this rests on your RTX 5090 run. Fine for a 21-line opt-in fallback.

#46 only touches alloc_encode_src_buffer; independent of #45/#47.

salty2011:
Both asks done, pushed:

* Safety wording scoped to LINEAR allocation failure, in the code comment and the PR body. Encode-time corruption (the GFX12 silent case) is called out as not covered.
* The `info!` now logs the effective tiling once, after allocation succeeds. Fallback hosts log `tiling=OPTIMAL` instead of four LINEAR lines plus four retractions.

Agreed on `vkGetPhysicalDeviceImageFormatProperties2` with the video profile chain as the correct probe. Keeping that as a follow-up so this stays a small fallback fix.

Build and fmt clean on `43d4c25`.

PR #47, opened 2026-07-26
fix(vulkan): replace guessed struct offsets with a header-compiled bridge; plug two leaks

salty2011:
The encode-src path reaches into GStreamer Vulkan structs by writing hardcoded byte offsets from Rust:

```rust
const BARRIER_QUEUE_OFFSET: usize = 296;      // barrier.parent.queue
const BARRIER_SEMAPHORE_OFFSET: usize = 320;  // barrier.parent.semaphore
const BARRIER_IMAGE_LAYOUT_OFFSET: usize = 368;
```

plus `#[repr(C)]` overlay structs mirroring `GstVulkanInstance`, `GstVulkanDevice`, `GstVulkanQueue` and `GstVulkanImageMemory`.

The comments record that the offsets were verified with an `offsetof` probe against the running headers, and they are correct today. The concern is the failure mode: if any of those structs gains or reorders a field, the result is a silent write to the wrong memory rather than a build error, and each overlay has to be re-verified by hand against every GStreamer version.

This replaces both mechanisms with a small C bridge (`vulkan_bridge.c`) compiled against the actual GStreamer Vulkan headers via `build.rs`, so field access is checked by the compiler and tracks the headers automatically. Two behavior changes ride along, both replacing guesses with failure:

* `raw_handles()` no longer falls back to queue family 0 when `select_queue(GRAPHICS)` returns null. It returns `None`, and `GsVulkanBuf::new` turns that into Vulkan output failure instead of submitting on a guessed family.
* `recover_vk_image()` returns `None` on a null image instead of handing back a garbage handle.

Both are intended. The layout and fan-out handling is unchanged.

### Two leaks surfaced while moving the barrier reset into the bridge

1. **Per-image `VkSemaphore` handle leak.** Clearing `barrier.parent.semaphore` by writing the field left the timeline semaphore the allocator created undestroyed — one leaked handle per encode-src image. The bridge now calls `vkDestroySemaphore` before clearing.

2. **Per-call `GstVulkanQueue` leak.** `gst_vulkan_device_select_queue()` is transfer-full, but `raw_handles()` only read the queue family off the returned object and dropped the pointer, leaking one queue object per call. It is now taken with `from_glib_full` and released once the family index has been read.

### Scope

Deliberately limited to the ABI-safety change and those two leaks. I also carry a separate change that makes the producer participate in `GstVulkanQueue`'s submission lock, which I have **not** included here: it was written for a `VK_ERROR_DEVICE_LOST` / Xid theory on my hardware that later turned out to have a different, proven cause (a buffer-reuse gate blinded by a header copy), so I have no evidence it fixes anything and did not want to present it as a fix. Happy to open it separately if you think it is correct on principle.

Build-verified on this branch's head (`43d4c25`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1858,6 +1858,7 @@ version = "0.2.0"
 dependencies = [
  "ash",
  "calloop",
+ "cc",
  "gstreamer",
  "gstreamer-allocators",
  "gstreamer-video",
--- a/ci/README.md
+++ b/ci/README.md
@@ -43,6 +43,7 @@ wherever `vulkanh264enc` registers (an nvidia render node, or Intel with
 
 - rust toolchain (`rustup`), `gst-launch-1.0` + gst-plugins-{base,bad,vaapi}
 - build deps: `libwayland-dev libinput-dev libxkbcommon-dev libgbm-dev libegl1-mesa-dev libudev-dev libclang-dev pkg-config`
+- a C compiler (`gcc`/`clang`) and the Vulkan loader headers (`libvulkan-dev`, or `vulkan-loader-devel` on Fedora) — `build.rs` compiles `vulkan_bridge.c` against the gst Vulkan headers
 - a Vulkan driver (`mesa-vulkan-drivers` for AMD/Intel, the proprietary driver for Nvidia) — the converter queries Vulkan for NV12 export modifiers; without it the source advertises nothing and negotiation fails
 - `hwdata` (for `/usr/share/hwdata/pci.ids`) so the GPU-name lookup test passes
 - Nvidia only: the `cuda` feature links `libgstcuda-1.0`; if your distro ships it without a `.pc`, add a `gstreamer-cuda-1.0.pc` (`Libs: -lgstcuda-1.0`) and a `libgstcuda-1.0.so` dev symlink
--- a/wayland-display-core/Cargo.toml
+++ b/wayland-display-core/Cargo.toml
@@ -45,6 +45,7 @@ libc = "0.2"
 
 [build-dependencies]
 pkg-config = "0.3"
+cc = "1"
 
 [dev-dependencies]
 tracing-subscriber = "0.3.18"
--- a/wayland-display-core/build.rs
+++ b/wayland-display-core/build.rs
@@ -1,6 +1,20 @@
-use pkg_config;
-
 fn main() {
+    // Compile a tiny, header-checked bridge for Vulkan handle/barrier fields that bindgen
+    // intentionally truncates because Vulkan handles are platform-dependent typedefs.
+    // This avoids guessing private struct offsets in Rust and makes any GStreamer ABI drift
+    // a compile-time failure against the exact headers used by the plugin.
+    let vulkan = pkg_config::Config::new()
+        .atleast_version("1.28")
+        .probe("gstreamer-vulkan-1.0")
+        .expect("gstreamer-vulkan-1.0 >= 1.28 is required");
+    let mut bridge = cc::Build::new();
+    bridge.file("src/utils/vulkan_bridge.c");
+    for include in vulkan.include_paths {
+        bridge.include(include);
+    }
+    bridge.compile("wayland_display_vulkan_bridge");
+    println!("cargo:rustc-link-lib=vulkan");
+
     // Check if the cuda feature is enabled
     #[cfg(feature = "cuda")]
     {
@@ -18,4 +32,5 @@ fn main() {
 
     // Rerun if build script changes
     println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=src/utils/vulkan_bridge.c");
 }
--- a/wayland-display-core/build.rs
+++ b/wayland-display-core/build.rs
@@ -13,7 +13,6 @@ fn main() {
         bridge.include(include);
     }
     bridge.compile("wayland_display_vulkan_bridge");
-    println!("cargo:rustc-link-lib=vulkan");
 
     // Check if the cuda feature is enabled
     #[cfg(feature = "cuda")]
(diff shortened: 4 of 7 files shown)


salty2011:
Housekeeping: this is independent of my other two PRs on this branch (#45 and #46). I verified all six pairwise orderings apply cleanly, and all three applied together build clean on `43d4c25` — so they can be merged in any order, individually, with no rebase needed. I split them rather than sending one PR so the small self-contained fixes are not held up by discussion of the larger one. Happy to combine them into a single PR if you would prefer one thread.

JBailes:
Validated both leaks against gst. _vk_image_mem_new_alloc_with_image_info creates the timeline semaphore at alloc, and _vk_image_mem_free only destroys it behind if (mem->barrier.parent.semaphore), so nulling the field from Rust meant that path never fired. Destroy-then-null is the right order. select_queue is transfer-full, so from_glib_full is right. The bridge beats the offset writes.

There is a third leak inside the leak-fixing function. image->barrier.parent.queue = NULL drops a strong ref; gst's own _vk_image_mem_free uses gst_clear_object (&mem->barrier.parent.queue). Harmless today because the only call site is straight after alloc where queue is NULL, but the function reads as general purpose. Use gst_clear_object.

The doc comment in vulkan_nv12.rs for "The exact GStreamer graphics queue used by queue; retained so every custom vkQueueSubmit participates in GStreamer's external-submit lock" has landed on outputs: Vec<Nv12Out>. It describes the submission-lock field from the change you excluded. #48 has the same class of drift, a doc comment landing on the wrong field of State. Please check for others.

"No behaviour change is intended" is not accurate. raw_handles used to fall back to family 0 when select_queue(GRAPHICS) returned null; now it returns None, and GsVulkanBuf::new turns that into total Vulkan output failure. recover_vk_image now returns None on a null image. Both are better than the old guesses, just state them.

The fan-out gate rewrite at vulkan_nv12.rs:905 is out of scope. Code is identical, but the rationale swaps from "our fence says nothing about the consumers' encode reads" to "vulkanh26x waits its GstVulkanOperation before returning", and the note explaining the 1s cap is gone. That is a different, unevidenced claim in a PR about struct offsets. Drop it here or evidence it separately.

Minor: the new pkg-config probe .expect()s while the CUDA probe below only warns; and without a shared .h, the extern decls and C definitions are still maintained independently. Smaller ABI surface, not zero.

Let's merge once the queue unref and the two comment issues are sorted.

salty2011:
All three sorted, pushed:

* `gst_clear_object (&image->barrier.parent.queue)` in the bridge, matching gst's own `_vk_image_mem_free`.
* Both drifted doc comments removed. The fan-out gate comment is restored to the original wording, 1s cap note included. `vulkan_nv12.rs` is now byte-identical to the base commit. I swept the rest of the diff for the same class of drift, nothing else.
* PR body now states the two behavior changes (no family 0 fallback, `recover_vk_image` returns None on null).

On the pkg-config `.expect()`: kept it. The bridge is required for this PR to work at all, so a warn-and-continue would only defer the failure to link time with a worse error. The CUDA probe can warn because that path is optional.

Agreed the extern decls and C definitions are still maintained by hand. Smaller surface, not zero.

JBailes:
@salty2011 Needs a rebase on your fork: base moved to 731bc70, and #45 and #46 both landed in the function this rewrites.

Watch one thing when you do it. Git puts #46's encode-src allocated with tiling={:?} log on the HEAD side, in the same hunk as the offset block you delete. Taking "theirs" wholesale drops it, and that log is the entire observable output of #46. Keep it, and take the bridge call for the offset writes only.

Both leaks check out against the 1.28.4 source. gstvkimagememory.c:169 creates the semaphore at alloc time and :400 destroys it only when non-NULL, so #37 did orphan one per image, and destroying before clearing is not a double free. gstvkdevice.c:1234 confirms select_queue is (transfer full). One correction: barrier.parent.queue is NULL at alloc (g_new0, never set in _init), so the gst_clear_object is a defensive no-op rather than a third leak.

The blocker for me is the libvulkan link. readelf -d gives base 0 NEEDED entries and this branch 1. Cargo.toml declares ash as features = ["loaded"] so a missing loader stays a runtime error; calling vkDestroySemaphore from the bridge forces cargo:rustc-link-lib=vulkan and reverses that, taking out the CUDA and VA paths that never wanted Vulkan. Go through vkGetDeviceProcAddr instead and drop the link-lib line.

Two small ones. ci/README.md:45 needs libvulkan-dev / vulkan-loader-devel and a C compiler added, since CI misses it and it lands on anyone building on a bare host. And wayland_display_vk_queue is never declared on the Rust side, so it can go.

I tested the rebased tree on my end. Renaming queue->family to a field that does not exist fails the build, which is the proof the bridge is genuinely header-checked. A vulkanh264enc pipeline returns 0 on an RX 7900 XTX with select_queue(GRAPHICS) non-null, so the stricter raw_handles gate does not bite there. What I cannot rule out is a device exposing no graphics queue where the old family-0 guess happened to work.

salty2011:
Rebased onto `9a8ced7`. All four addressed.

* **libvulkan link fixed.** `vkDestroySemaphore` now resolves via `gst_vulkan_device_get_proc_address`, and the `link-lib` line is gone. `readelf -d` before: `libgstvulkan-1.0.so.0` + `libvulkan.so.1`. After: `libgstvulkan-1.0.so.0` only, matching base exactly.
* **#46's tiling log kept.** Thanks for the heads-up, that hunk resolves badly. Log preserved, offset writes gone, and the fan-out rationale the offset block carried is now on the bridge call.
* **Your correction taken.** `barrier.parent.queue` is NULL at alloc, so the `gst_clear_object` is a defensive no-op, not a third leak. Kept with a comment saying exactly that, since it states the postcondition and matches gst's own `_vk_image_mem_free`. Happy to drop it if you'd rather.
* `wayland_display_vk_queue` deleted. `ci/README.md` now lists a C compiler and the Vulkan loader headers.

`cargo fmt` clean, `check` and `check --features cuda` clean, 23 tests pass.

PR #48, opened 2026-07-26
feat(vulkan): own the GstVulkanInstance/Device per element instead of process-globals

salty2011:
## Problem

The owned `GstVulkanInstance` + `GstVulkanDevice` live in two process-global `OnceLock` slots, so every `waylanddisplaysrc` in a process shares whichever one was minted first. That means:

- Session 2..N inherit session 1's `target_minor`, ignoring their own.
- One session's `DEVICE_LOST` corrupts all N.
- Nothing ever hands the device back. The slot leaks by design.

## Change

- One `VulkanShare` per element, cloned once into that element's compositor thread. Same pattern as the existing `app_surface_commits` / `renderer_degraded` `Arc`s.
- Each element mints its own device and answers `gst.vulkan.{instance,device}` with it. Ownership alone retires it: the `Arc` drops when the element is finalized, so N sessions no longer leak N devices.
- Storage location only. `ensure_owned_device` is byte-identical to the old global path, no vendor branch, RADV unchanged. C bridge untouched.

Rebased onto `b49f7c0` as a single commit now that #46 and #47 have landed.

## Your review points

**Fan-out topology.** Inside this crate, every buffer is minted on its own element's device (`set_caps` -> `ensure_owned_device`, ring built from the same share). Which device the *encoder* adopts is the embedder's context wiring, since the encoder is in a separate pipeline across interpipe.

`handle_set_context` now keeps the device this element already holds and warns if a different one is offered. Your bin-fanout path is what convinced me: GstBin fans a have-context to every child, so two sources in one bin is enough, no mis-wired embedder needed. The cross-device pairing is refused in the crate rather than left to the embedder to check.

Ordering contract worth stating: inject a device before caps negotiation or not at all. `set_caps` mints eagerly, so a later injection is refused.

**stop()/start().** `clear()` is gone. Per-element ownership plus the compositor thread join already retire everything that matters at `stop()`, so it only made release earlier, and it was the sole source of the desync caveat this PR previously pushed onto embedders. What remains after `stop()` is a bare `VkDevice`+`VkInstance` held until the element is finalized.

**Feature-gate.** I believe this is already true on base, so I've left it: `Cargo.toml` declares `gstreamer-vulkan` `features = ["v1_28"]` unconditionally, and `utils/mod.rs` declares `vulkan_share`/`vulkan_nv12` unconditionally. A CUDA-only gst-1.24 build fails before this PR. The only new requirement was headers + a C compiler, documented in `ci/README.md` by #47. Say the word if you still want a `vulkan` feature.

**target_minor after an absorbed device.** Left as-is, documented. The slot records no provenance, so the short-circuit cannot tell a self-minted device from an injected one. When they differ the encoder is already bound to the injected device, and re-minting on the requested node would split producer and encoder across two `VkDevice`s, which is worse than honouring the wrong node.

**Also done:** tautological test dropped · `repr(C)` overlay doc claim fixed · `comp/mod.rs` doc drift fixed · lock order documented on `ensure_owned_device` (only method holding both, so the inverse order can't arise) · `lib.rs` five-second comment deleted, that path cannot produce a VULKAN video info.

**Not needed:** `new_ret_no_self` doesn't fire (clippy exempts `new` returning `Arc<Self>`) · `raw_handles` is byte-identical to base, so the family-0 point is moot.

## Verified

`cargo fmt` clean · `check` and `check --features cuda` clean · rustdoc no broken links · 23 tests pass.

### DIFF
--- a/gst-plugin-wayland-display/src/waylandsrc/imp.rs
+++ b/gst-plugin-wayland-display/src/waylandsrc/imp.rs
@@ -47,6 +47,14 @@ pub struct WaylandDisplaySrc {
     /// `caps()` falls back to the hardcoded `HDR_MASTERING` / `HDR_CLL` defaults. Driven in
     /// `create()`, read in `caps()`. Never consulted when WOLF_HDR_CM is unset.
     hdr_meta: Mutex<(Option<String>, Option<String>)>,
+    /// This element's OWN Vulkan-encode device share, replacing the process-global
+    /// `vulkan_share` slots so N concurrent Vulkan sessions in one process each mint, own and
+    /// destroy their own `VkDevice` (an isolated failure domain). Lives for the element's whole
+    /// lifetime (created in `Default`) so `set_context`, the context query and `set_caps` can
+    /// reach it regardless of start/stop ordering; a clone goes to this element's compositor
+    /// thread in `start()`. The device is retired by ownership: when the element is finalized
+    /// its `Arc` drops, so N sessions do not leak N devices.
+    vulkan_share: Arc<waylanddisplaycore::utils::vulkan_share::VulkanShare>,
 }
 
 impl Default for WaylandDisplaySrc {
@@ -59,6 +67,7 @@ impl Default for WaylandDisplaySrc {
             command_rx: Mutex::new(Some(command_rx)),
             hdr_active: AtomicBool::new(false),
             hdr_meta: Mutex::new((None, None)),
+            vulkan_share: waylanddisplaycore::utils::vulkan_share::VulkanShare::new(),
         }
     }
 }
@@ -569,9 +578,9 @@ impl ElementImpl for WaylandDisplaySrc {
     fn set_context(&self, context: &Context) {
         // Absorb a downstream encoder's shared GstVulkanDevice so the Vulkan-encode path can
         // mint encode-src images on the same device (best-effort; no-op for other contexts).
-        if waylanddisplaycore::utils::vulkan_share::handle_set_context(context) {
-            tracing::info!("waylandsrc: absorbed shared GstVulkanDevice for Vulkan encode path");
-        }
+        // Returns true whenever a device is shared afterwards, including when the guard kept
+        // the one this element already had; vulkan_share logs which of the two happened.
+        self.vulkan_share.handle_set_context(context);
 
         // Absorb a downstream VA encoder's GstVaDisplay context so our NV12 buffers can
         // attach VA surfaces on the same display (best-effort).
@@ -705,11 +714,8 @@ impl WaylandDisplaySrc {
         }
         let node = render_node.unwrap_or_else(|| "/dev/dri/renderD128".into());
         let minor = waylanddisplaycore::utils::vulkan_nv12::render_node_minor(&node);
-        waylanddisplaycore::utils::vulkan_share::provide_context(
-            self.obj().upcast_ref::<gst::Element>(),
-            query,
-            minor,
-        )
+        self.vulkan_share
+            .provide_context(self.obj().upcast_ref::<gst::Element>(), query, minor)
     }
 }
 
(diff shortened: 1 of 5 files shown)


JBailes:
The diagnosis is right: process-global slots mean session 2..N inherit session 1's target_minor and its DEVICE_LOST, and nothing hands a device back.

Fan-out is the question I want answered. This code exists because one buffer goes to N encoders. A buffer allocated on element A's VkDevice reaching an encoder that adopted element B's is invalid, not slow. You would know better than me whether Wolf's graph can produce that pairing. If it can, the share wants to hang off the device, which is the alternative you already offered.

stop() then start() can desync producer from encoder. clear() drops the device, start() mints a fresh one, and an encoder that already adopted the old one may not re-query. The global was stable across restarts. Worth a line either way.

The build now hard-requires gst-vulkan. .expect("gstreamer-vulkan-1.0 >= 1.28") is unconditional and the bridge needs a header with VK_IMAGE_LAYOUT_VIDEO_ENCODE_SRC_KHR. A CUDA-only build on gst 1.24 stops building. Feature-gate it, and take link flags from probe() rather than hardcoding -lvulkan with no search path.

Three rebase leaks, all outside scope. The comp/mod.rs doc block landed inside hdr_candidate_since's comment. A doc about a retained graphics queue is attached to outputs: Vec<Nv12Out>, and no such field exists here. And the fan-out rationale in vulkan_nv12.rs is rewritten from "consumers' encode reads have no shared sync with us" to "writability is the completion gate", with the ~1s cap note deleted and no code change. That reverses a #37 hazard claim, and if you mean it, it needs its own PR.

The test passes against a broken implementation. Two fresh Arcs being distinct and empty holds however the storage works. Its doc comment is cut mid-sentence. I would rather drop it than count it as coverage.

Smaller: the VulkanShare doc says the repr(C) overlays are untouched and this PR deletes them; new() -> Arc<Self> trips new_ret_no_self; ensure_owned_device holds the device lock across instance open, Entry::load, enumeration and device open, taking the instance lock inside it, opposite to provide_context; raw_handles no longer falls back to family 0, which is probably better and is not called out; and the WaylandDisplay::new path gets a share nothing can fill, so apply_video_info waits the full five seconds before degrading.

Land #46 and #47 on their own, then bring this back as the single commit with the fan-out question answered in the body. The failure modes I care about are GPU hangs and cross-device handles, and unit tests reach neither, so I want your answer on topology more than I want more tests.

salty2011:
Rebased onto `b49f7c0` as a single commit now that #46 and #47 are in. Fan-out topology answered in the body, along with two of your points I think are already resolved on base. Details there rather than here.

JBailes:
Okay. This looks a lot better.

handle_set_context still overwrites unconditionally (vulkan_share.rs:103), ungated by settings.vulkan, unlike the query path. Under the global that was free. Everyone shared one VkDevice. Per-element it is the hazard you named, reachable without a mis-wired embedder: two waylanddisplaysrc in one bin, B's encoder posts have-context, GstBin fans it to every child, A's share becomes B's device while A's encoder stays on A's. Fix is three lines in the same place: if self.device is Some and the incoming pointer differs, warn and keep ours. That also puts the pointer-equality check Wolf performs inside the crate, where an embedder cannot forget it.

And one question. clear() in stop() is justified by the leak, but per-element ownership plus the thread join already fixes the leak. clear() only makes release earlier, and it is the sole source of the restart desync he pushes onto embedders. Does anything hold a stopped element alive long enough for early release to matter? If not, delete it.

Smaller: ensure_owned_device short-circuits on an absorbed foreign device, so session N's own target_minor is still silently ignored when an embedder injects first. The lib.rs five-second comment describes something unreachable, since capi.rs:105 cannot produce GstVideoInfo::VULKAN.

salty2011:
All four addressed.

* **`handle_set_context` guard added.** Keeps this element's device, warns on a different one. Your bin-fanout path is the one that convinced me, it needs no mis-wired embedder.
* **`clear()` deleted.** You were right that ownership plus the thread join already do the work. Worth noting the restart argument is moot either way: `start()` takes the command channel with `take().unwrap()` and it is never replenished, so a second `start()` panics. Pre-existing, not this PR.
* **`target_minor`**: left as-is, documented. The slot records no provenance, so the short-circuit cannot tell a self-minted device from an injected one; when they differ the encoder is already bound and re-minting would split producer and encoder across devices.
* **`lib.rs` five-second comment**: deleted, you are right that path cannot produce a VULKAN video info.

JBailes:
Merged to vulkan

PR #49, opened 2026-07-26
fix(vulkan): complete the GLES render fence before the Vulkan conversion reads it

salty2011:
The NV12 / VulkanImage output paths render the scene into a GLES RGBA dmabuf and then import that same allocation into Vulkan for the RGBA→NV12 conversion.

The render fence is waited in the caller, **after** `State::create_frame()` returns — but `create_frame()` has already called `to_gs_buffer()`, which submits the Vulkan read. So the wait only orders the later GStreamer hand-off, not the GLES→Vulkan ownership seam: the GLES write and the Vulkan read end up ordered only if the driver happens to serialize the two APIs, which nothing guarantees.

This moves the existing wait to immediately after `render_output()`, before any readback or `to_gs_buffer()` conversion, and drops the now-redundant late wait from the caller.

**No new synchronization is introduced** — the same fence is waited, just at the point where it actually protects the consumer. Net diff is 2 files, +11/−4.

Independent of the encoder and of any downstream element: it concerns only the GLES-producer / Vulkan-consumer seam inside the compositor.

### On attribution

I found this while chasing NVIDIA device-loss reports (Xid 13/32 → `VK_ERROR_DEVICE_LOST`) on an RTX 5090. To be straight with you: **that symptom turned out to have a different, separately proven cause**, so I am not claiming this fixes device loss. It is offered as a correctness/ordering fix on its own merits — the current code depends on undocumented driver behaviour for the seam to be safe.

### Verification

`cargo build --release` clean and `cargo test -p wayland-display-core` green (18 passed, 0 failed) on this branch's head; `cargo fmt --check` clean. This fix has been running in my deployment on both NVIDIA and AMD hosts.

### DIFF
--- a/wayland-display-core/src/comp/mod.rs
+++ b/wayland-display-core/src/comp/mod.rs
@@ -836,10 +836,6 @@ pub(crate) fn init(
                         }
                         if let Err(_) = match state.create_frame() {
                             Ok((buf, render_result)) => {
-                                render_result
-                                    .sync
-                                    .wait()
-                                    .expect("Error during render_result.sync"); // we need to wait before giving a hardware buffer to gstreamer or we might not be done writing to it
                                 let res = buffer_sender.send(Ok(buf));
                                 let rendered_states = &render_result.states;
                                 let rendered_damage = render_result.damage.is_some();
--- a/wayland-display-core/src/comp/rendering.rs
+++ b/wayland-display-core/src/comp/rendering.rs
@@ -221,6 +221,17 @@ impl State {
             [0.0, 0.0, 0.0, 1.0],
         )?;
 
+        // The NV12/VulkanImage paths import this GLES render target as external Vulkan
+        // memory and sample it during `to_gs_buffer`. Waiting in the caller after
+        // `create_frame` returns is too late: the converter has already submitted its
+        // Vulkan read by then, so the GLES write and the Vulkan read are only ordered if
+        // the driver happens to serialize the two APIs -- which nothing guarantees.
+        // Complete the GLES render here, before any hardware conversion or hand-off
+        // consumes the target.
+        render_output_result
+            .sync
+            .wait()
+            .expect("Error during render_result.sync");
         // HDR render-path spike ISOLATION probe: read the fp16 RGBA dmabuf straight off the GLES
         // framebuffer (after render_output, before the Vulkan converter in to_gs_buffer) and log
         // the bar-center float values -- proving whether GLES kept >1.0 or clamped. Gated behind


JBailes:
Yo @salty2011,

I checked your premise out, and it seems right. create_frame (rendering.rs:212) falls into to_gs_buffer (allocator/mod.rs:729), which calls v.convert on the RGBA dmabuf. convert_inner (vulkan_nv12.rs:888) never touches a fence on that input. The only sync_file traffic is IMPORT_SYNC_FILE on the output (:1443). The old caller-side wait was too late.

The problem is with ExportFence. GlesFrame::finish_internal (gles/mod.rs:2309) returns a real fence only when Capability::ExportFence is present. Otherwise it calls glFinish and returns SyncPoint::signaled(), and wait() on an empty SyncPoint does nothing (sync/mod.rs:85). One path was already safe and this wait is a no-op there. On the other, there is a race. "runs fine in my deployment" does not tell us which.

This is a CPU stall doing a semaphore's job. The wait now sits between render_output and to_gs_buffer, so the CPU cannot record and submit the conversion while the GPU draws. SyncPoint::export() (sync/mod.rs:104) hands back a sync_file fd. vulkan_nv12.rs already loads khr::external_semaphore_fd (:568) and creates export semaphores (:692). Import it and wait at COMPUTE_SHADER on the conversion submit. Nobody has measured the stall.

The reverse problem is still open. On implicit_sync, convert_inner submits and returns without waiting the fence (:1177), so the compute read is still in flight. The compositor renders into one stable RGBA buffer every frame (:890). The next frame's GLES write races the previous frame's Vulkan read. Vulkan does no implicit dmabuf sync and nothing attaches a read fence to the input BO, so Mesa has nothing to wait on. NVIDIA is safe because wait_for_fences blocks (:1187). Same seam, same buffer, same argument you make in the PR.

salty2011:
Ran your questions down on the RTX 5090 host:

* **Which path:** `EGL_ANDROID_native_fence_sync` is present (confirmed in the EGL display extension dump), so `finish_internal` returns a real exportable fence and the wait is live, not the glFinish no-op. The race this PR orders was real on this deployment.
* **Stall cost, measured:** timed the moved `sync.wait()` over 3900 frames at 1080p60 under continuous damage. Steady state averages ~145 us per frame, max 3.5 ms (startup outliers only). Under 1% of the frame budget.
* **Semaphore import:** agreed it is the right endpoint, and the infra is already loaded (`khr::external_semaphore_fd` at vulkan_nv12.rs:568). Given the measured stall is ~145 us I would rather land this ordering fix as-is and do the `SyncPoint::export()` to `COMPUTE_SHADER`-wait import as a follow-up, so the correctness fix is not held on new sync plumbing. Happy to do it in this PR instead if you prefer.
* **Reverse race (GLES write vs in-flight Vulkan read on implicit_sync):** agreed it exists and this PR does not close it. It is the same seam on the other side, AMD-only (`wait_for_fences` covers NVIDIA). I would take that as a separate PR since it needs a read-fence attached to the input BO, which is new mechanism, not ordering.

JBailes:
Merged.

@salty2011 , for the future: GPU wait instead of a CPU one. SyncPoint::export() -> Option exists at the pinned smithay rev (sync/mod.rs:104), with is_exportable() for the fallback, and the converter already builds a khr::external_semaphore_fd loader for the export direction. Import the GLES fence as a wait semaphore on the convert submit and you get the 386 us back. Follow-up, not a blocker.

Two nits. The panic message still reads "Error during render_result.sync", naming the caller's old variable rather than render_output_result. And the deleted comment carried a second reason, that a hardware buffer must not reach GStreamer while it is still being written; the new placement still satisfies it, but that half is now unstated and it is the part covering the plain-dmabuf path.

PR #50, opened 2026-07-26
fix(egl): stop reporting a failed wl_display bind as lost hardware acceleration

salty2011:
One small logging fix around the EGL `wl_display` bind.

`renderer.bind_wl_display()` failing is logged as `info!("Failed to initialize EGL hardware-acceleration")`. That reads as a fallback to software rendering, but the only product of this bind is the `EGLBufferReader` used by the legacy wl_drm / EGL-image import path. Both client-buffer routes here go through `import_dmabuf` (`handlers/dmabuf.rs`, and `handlers/wl_drm.rs`, since mesa's wl_drm is implemented over dmabuf), so a failed bind disables neither hardware acceleration nor dmabuf import.

On render nodes where the bind always fails, that line appears on every startup. It cost me a long detour chasing a software-rendering fallback that was never happening. Demoted to `debug!` and reworded to say what is actually unavailable.

The earlier revision of this PR also added an explicit `unbind_wl_display()` at teardown. Review showed the mechanism I proposed for it was wrong, and a live reproduction confirmed the real cause is overlapping compositor lifetimes, which prompt unbinding does not address. Dropped; details in the comments.

`cargo build` and `cargo fmt --check` clean on `43d4c25`.

### DIFF
--- a/wayland-display-core/src/comp/mod.rs
+++ b/wayland-display-core/src/comp/mod.rs
@@ -322,9 +322,19 @@ impl State {
                 dmabuf_state.create_global::<State>(&dh, formats.clone())
             };
 
+            // The only product of this bind is the `EGLBufferReader` used by the legacy
+            // wl_drm / EGL-image import path. Both client-buffer routes here go through
+            // `import_dmabuf` (`handlers/dmabuf.rs`, and `handlers/wl_drm.rs` -- mesa's wl_drm
+            // is implemented over dmabuf), so a failed bind does not disable hardware
+            // acceleration and does not affect dmabuf import. Logging it as
+            // "Failed to initialize EGL hardware-acceleration" is misleading: on render nodes
+            // where the bind always fails it reads as a fallback to software rendering.
             match renderer.bind_wl_display(&dh) {
                 Ok(_) => tracing::info!("EGL hardware-acceleration enabled"),
-                Err(err) => tracing::info!(?err, "Failed to initialize EGL hardware-acceleration"),
+                Err(err) => tracing::debug!(
+                    ?err,
+                    "EGL wl_display bind unavailable (legacy wl_drm/EGL-image import disabled; dmabuf import unaffected)"
+                ),
             }
 
             // wl_drm (mesa protocol, so we don't need EGL_WL_bind_display)


JBailes:
@salty2011 

I agree with you on issue 1. Issue 2 doesn't make sense, the mechanism doesn't match the source. Three things:

No shared Arc. GlesRenderer owns the reader outright: egl_reader: Option<EGLBufferReader> (gles/mod.rs:289). The only Arc is inside it, wayland: Option<Arc<*mut wl_display>> (egl/display.rs:1011), created once in EGLBufferReader::new (:1038) and never cloned. WeakBufferReader::upgrade() (:1021) builds readers with wayland: None precisely so they can't unbind. Arc::try_unwrap in the Drop impl (:1219) therefore always succeeds, and the unbind is prompt.

Teardown isn't async. WaylandDisplay::drop (lib.rs:346-354) sends Quit, then joins the compositor thread. init has returned and state has dropped before the next WaylandDisplay exists.

The drop already does it. Drop for GlesRenderer (gles/mod.rs:1777) takes egl_reader with the context current. If make_current() fails and that line is skipped, egl_reader is still a struct field and drops right after Drop::drop returns.

So the new call runs microseconds ahead of a drop that did the same thing.

It's also slightly less robust. comp::init is wrapped in catch_unwind (lib.rs:148) and is full of .expect()/.unwrap(), including in the event-loop callback. A panic skips the explicit call at the end of init but still runs state's destructor. The implicit path is panic-safe; this one isn't.

What I think you probably hit: OtherEGLDisplayAlreadyBound wraps a real eglBindWaylandDisplayWL failure (egl/display.rs:855), and an EGLDisplay only takes one wl_display at a time. With join-on-drop, slow teardown doesn't explain it. Overlapping lifetimes do: two WaylandDisplay instances live on the same render node at once. That's normal in GStreamer, where the new element is set up before the old one is released, and this branch has interpipe fan-out. Unbinding promptly does nothing for an overlap.

Worth checking the logs for the "Double bind_wl_display" warning (:860). It keys off a Weak<EGLDisplayHandle> upgrade rather than the wl_display, so if it appears next to the error, that confirms overlap.

salty2011:
You called it. I reproduced the overlap on live hardware.

Five back-to-back session cycles on an RTX 5090 host (new compositor launched while the previous one tears down). One cycle logged, during the NEW compositor's init:

```
WARN smithay::backend::egl::display: Double bind_wl_display, smithay does not support this, please report
INFO waylanddisplaycore::comp: EGL hardware-acceleration enabled
```

So two `WaylandDisplay` lifetimes do overlap on the same render node in this deployment (session teardown drains while the next session starts), and the bind succeeded through the weak-upgrade path that time. My `OtherEGLDisplayAlreadyBound` was the same overlap losing the race, not a stale bind from slow teardown. Prompt unbinding cannot help an overlap, so part 2 is dropped.

The PR is now just the logging demotion, +11/-1. Body updated.

JBailes:
@salty2011 Based on the latest changes, I am closing this and creating a new PR for the intent.

PR #51, opened 2026-07-27
fix(vulkan): preserve image flags during allocation

elibosley:
## What changed

- add a GStreamer patch that forwards `VkImageCreateInfo::flags` to
  `vkGetPhysicalDeviceImageFormatProperties()`
- apply the patch in the Vulkan runtime image
- document why the patch is required and when it can be removed

## Root cause

`gst_vulkan_image_memory_alloc_with_image_info()` creates the image with the
caller's flags, but its post-create format query currently passes `0`. On RADV,
that makes an otherwise valid multiplanar NV12/P010 allocation fail when
`MUTABLE_FORMAT` and `EXTENDED_USAGE` authorize storage-capable plane views.
The wrapped-image allocation path already preserves these flags.

## Impact

Compatible Vulkan Video implementations can write conversion output directly
into the encoder-owned NV12/P010 image instead of falling back to a separate
storage scratch image and copy. Implementations that do not advertise the
required format/usage combination retain the existing fallback.

## Validation

- patch applies cleanly to GStreamer 1.28.4 and current GStreamer `main`
- the patched GStreamer/Wolf images built successfully with the protected
  Vulkan, HDR, shared-frame ownership, and input-idempotency tests
- live RADV probes on an AMD Radeon 780M at 3840×2160 selected
  `direct_storage=true`, optimal tiling, for both NV12/H.264 and P010/H.265
- local 3840×2160/60 DMA-BUF presentation remained zero-copy

The same minimal source change is prepared against GStreamer `main`; this
repository patch can be removed once an upstream release contains it.

### DIFF
--- a/docker/vulkan.Dockerfile
+++ b/docker/vulkan.Dockerfile
@@ -55,11 +55,13 @@ RUN dnf install -y \
 # --- Patched GStreamer 1.28.4 -> /opt/gst -----------------------------------
 COPY patches/vkh264enc-dpb-pool-in-new-sequence.patch /tmp/dpb.patch
 COPY patches/vulkanh265enc.patch /tmp/h265.patch
+COPY patches/gstvulkan-preserve-image-create-flags.patch /tmp/image-flags.patch
 RUN git clone --depth 1 --branch ${GST_VERSION} \
       https://gitlab.freedesktop.org/gstreamer/gstreamer.git /tmp/gstreamer && \
     cd /tmp/gstreamer && \
     git apply /tmp/dpb.patch && \
     git apply /tmp/h265.patch && \
+    git apply /tmp/image-flags.patch && \
     # auto_features=disabled leaves several subprojects' docs/meson.build referring
     # to an undefined plugins_cache_generator; short-circuit each when doc is off.
     for d in subprojects/*/docs/meson.build docs/meson.build; do \
@@ -86,7 +88,7 @@ RUN git clone --depth 1 --branch ${GST_VERSION} \
       -Dnls=disabled -Dgst-examples=disabled -Drs=disabled && \
     meson compile -C build && \
     meson install -C build && \
-    rm -rf /tmp/gstreamer /tmp/dpb.patch /tmp/h265.patch
+    rm -rf /tmp/gstreamer /tmp/dpb.patch /tmp/h265.patch /tmp/image-flags.patch
 
 ENV PKG_CONFIG_PATH=/opt/gst/lib64/pkgconfig \
     LD_LIBRARY_PATH=/opt/gst/lib64 \
--- a/patches/README.md
+++ b/patches/README.md
@@ -6,6 +6,7 @@ Apply against a gstreamer monorepo checkout before building:
 ```
 git apply patches/vkh264enc-dpb-pool-in-new-sequence.patch
 git apply patches/vulkanh265enc.patch
+git apply patches/gstvulkan-preserve-image-create-flags.patch
 ```
 
 ## vkh264enc-dpb-pool-in-new-sequence.patch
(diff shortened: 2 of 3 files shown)


salty2011:
Nice find. Read the allocator myself, bug is real: the image gets built with the caller's flags, then the format query passes `0`, so an image `vkCreateImage` just accepted can still be rejected.

One nit on the description:

- "the wrapped-image path already preserves these flags" is true on GStreamer `main` (`_vk_image_mem_new_wrapped_with_image_info` passes `image_info->flags`).
- On `1.28`, which this Dockerfile pins, the wrapped path goes through `_create_info_from_args()` and hardcodes `.flags = 0`, and its query passes a literal `0` too.
- Fix is correct either way. Just worth naming the branch so nobody hunts for that code in 1.28.

Two questions so I can reproduce:

**1. Which tree did the validation run on?**
- No `direct_storage` anywhere on `feat/vulkan-encode`.
- No tests matching "protected Vulkan", "shared frame ownership", or HDR.
- Closest idempotency test is `test_resolution.rs::idempotent_same_video_info_is_safe_and_stays_on_mode`, which is resolution, not input.
- If that all lives on a downstream fork, no problem, I would just like to know where to look.

**2. Is there a companion change that sets `MUTABLE_FORMAT` and `EXTENDED_USAGE`?**
Nothing on this branch passes non-zero flags into that function today:
- `alloc_encode_src_buffer()` in `vulkan_share.rs` never calls `.flags()`, so `image_info->flags` is `0` and the patch is a no-op there.
- `GstVulkanImageBufferPool` only sets those bits when usage is sampleable and not a video usage, which rules out the `VIDEO_ENCODE_SRC` pool.
- The NV12 conversion images are allocated straight through `ash` and never reach this allocator.

So this is groundwork rather than a behavior change, which is fine. Where does the direct write land? That is the part that drops the scratch plus copy, and it is the real win here.

Happy to test. I run Vulkan encode on both an RTX 5090 and AMD, so can cover NVIDIA as well as RADV. Worth doing, since passing the real flags makes the query stricter and an alloc that passed before could start failing on a driver that does not advertise that flag combination.

JBailes:
@elibosley I'm all for merging this, but please address @salty2011 's 3 issues.
I see a few issues in some of the statements being made, but none of them fundamentally impact the code itself outside of the 3 issues Salty raised, and the code is correct and good to go once Salty's issues are fixed.

elibosley:
> @elibosley I'm all for merging this, but please address @salty2011 's 3 issues. I see a few issues in some of the statements being made, but none of them fundamentally impact the code itself outside of the 3 issues Salty raised, and the code is correct and good to go once Salty's issues are fixed.

Happy to do so, sorry, my AI sloppified the above comment and made this very confusing - I've been working on a way to do local output through Gamescope within wolf, which if your team is interested in I can absolutely make some PRs to share, I love the way wolf uses lobbies etc. and I wanted to be able to use a direct output mechanism, so I've been working on this for ~2 weeks, but kept the code local until it's ready to share / upstream (or not if you want to avoid this complexity).

JBailes:
> > @elibosley I'm all for merging this, but please address @salty2011 's 3 issues. I see a few issues in some of the statements being made, but none of them fundamentally impact the code itself outside of the 3 issues Salty raised, and the code is correct and good to go once Salty's issues are fixed.
> 
> Happy to do so, sorry, my AI sloppified the above comment and made this very confusing - I've been working on a way to do local output through Gamescope within wolf, which if your team is interested in I can absolutely make some PRs to share, I love the way wolf uses lobbies etc. and I wanted to be able to use a direct output mechanism, so I've been working on this for ~2 weeks, but kept the code local until it's ready to public.

I'd certainly be interested in this, as it aligns with a project I'm doing, but I'm not sure it aligns with where @ABeltramo wants to go. I'd ask @ABeltramo for his take on it!

elibosley:
Here's the other PR: https://github.com/games-on-whales/gst-wayland-display/pull/52

FWIW I've tested on Nvidia and AMD but I likely have _not_ done enough testing or validation for me to feel completely comfortable with this yet (#52). I feel confident that this bug is going to restrict #52 so it would be good to get this upstreamed regardless.

salty2011:
@elibosley when you say local output do you mean to an attached display?

JBailes:
Merging, but with the note the writeup is not valid/relevant.

JBailes:
Actually, CI failure. I can't fix the CI failure due to it being in your fork @elibosley , could you fix it please?

PR #52, opened 2026-07-30
feat(vulkan): write directly into compatible encode images

elibosley:
## Summary
The shared Vulkan encoder path can now write NV12 directly into a compatible encode image, eliminating the scratch image and copy on drivers that support the required storage and video-encode combination.

## Why This Exists
The shared encoder image was already created with `VIDEO_ENCODE_SRC`, but compute conversion still wrote to a separate storage image and copied the result into that encoder image. This added an allocation, synchronization, and full-frame transfer to every compatible frame.

## Resolution
Probe `STORAGE | VIDEO_ENCODE_SRC` support with the active video profile, image-format list, tiling, and creation flags. When supported, allocate the encoder image with storage usage and create plane views directly on it. When unsupported or allocation fails, retain the existing scratch-image-and-copy path.

## Reviewer Considerations
- The direct capability probe deliberately includes the same video profile, format list, usage, and create flags as allocation so it does not assume storage support from the video-encode capability alone.
- The bundled GStreamer 1.28.4 patch is required because its allocator validated image format properties with zero flags even when the actual image used non-zero flags. It covers only the allocation-with-image-info path used here; the legacy wrapped-image helper remains unchanged.
- Hardware coverage is important: drivers may reject the requested flag/usage/profile combination, in which case the code must stay on the existing fallback path.

## Behavior Changes
- Supported Vulkan video-encode configurations convert directly into the encode source image.
- Unsupported configurations and allocation failures keep the portable scratch-and-copy behavior.

## Implementation Summary
- Add a GStreamer 1.28.4 allocator backport that preserves image creation flags during format validation.
- Query and allocate direct storage-capable encode images with `MUTABLE_FORMAT` and `EXTENDED_USAGE`.
- Use plane storage views of the encode image for direct conversion, with the existing scratch image retained as fallback.
- Add static usage/flag tests, document the GStreamer patch, and force a fresh Docker plugin build.

## Verification
- `cargo fmt --check` — passed.
- `git diff --check origin/feat/vulkan-encode...HEAD` — passed.
- `cargo test -p wayland-display-core direct_encode --lib` — blocked before compilation in this macOS checkout because `wayland-server.pc` is unavailable; the container image runs `cargo test --release -p wayland-display-core direct_encode`.

## Risk
Moderate, hardware-specific: a driver can decline the combined image capabilities. The capability probe and allocation-failure fallback preserve the prior scratch-and-copy path; NVIDIA and RADV encode smoke tests are still needed.

### DIFF
--- a/docker/vulkan.Dockerfile
+++ b/docker/vulkan.Dockerfile
@@ -55,11 +55,13 @@ RUN dnf install -y \
 # --- Patched GStreamer 1.28.4 -> /opt/gst -----------------------------------
 COPY patches/vkh264enc-dpb-pool-in-new-sequence.patch /tmp/dpb.patch
 COPY patches/vulkanh265enc.patch /tmp/h265.patch
+COPY patches/vulkan-image-create-flags-1.28.patch /tmp/image-create-flags.patch
 RUN git clone --depth 1 --branch ${GST_VERSION} \
       https://gitlab.freedesktop.org/gstreamer/gstreamer.git /tmp/gstreamer && \
     cd /tmp/gstreamer && \
     git apply /tmp/dpb.patch && \
     git apply /tmp/h265.patch && \
+    git apply /tmp/image-create-flags.patch && \
     # auto_features=disabled leaves several subprojects' docs/meson.build referring
     # to an undefined plugins_cache_generator; short-circuit each when doc is off.
     for d in subprojects/*/docs/meson.build docs/meson.build; do \
@@ -86,7 +88,7 @@ RUN git clone --depth 1 --branch ${GST_VERSION} \
       -Dnls=disabled -Dgst-examples=disabled -Drs=disabled && \
     meson compile -C build && \
     meson install -C build && \
-    rm -rf /tmp/gstreamer /tmp/dpb.patch /tmp/h265.patch
+    rm -rf /tmp/gstreamer /tmp/dpb.patch /tmp/h265.patch /tmp/image-create-flags.patch
 
 ENV PKG_CONFIG_PATH=/opt/gst/lib64/pkgconfig \
     LD_LIBRARY_PATH=/opt/gst/lib64 \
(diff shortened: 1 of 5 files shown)


elibosley:
I would say this is not 100% ready but I'd love some testing to occur on this work, it does allow direct allocation and has been tested on nvidia and RADV but I want more eyes.

salty2011:
Not a problem I'm able to give this a test on both Nvidia and AMD over the weekend.

Any specific testing I should do?

elibosley:
My robot whipped this up, sorry it's so prescriptive 😆 

## Hardware test checklist

The unit tests cover the requested Vulkan flags and usages. Before this leaves draft, please run the following on both NVIDIA and RADV video encode:

1. Build the Vulkan container. This also runs `cargo test --release -p wayland-display-core direct_encode`.
2. Enable the plugin's debug tracing by adding this environment variable to the Wolf/container process, then restart it:

   ```bash
   GST_DEBUG=waylanddisplaysrc:5
   ```

   The plugin forwards Rust tracing through GStreamer's `waylanddisplaysrc` category, so `RUST_LOG` alone will not show these messages. Use the service/container logs to inspect the result.
3. Start a real H.264 stream; also test HEVC and P010/HDR when available. Cover 1080p and 4K, then start/stop or reconnect a few times.
4. Confirm which route was selected in the logs:
   - `direct_storage=true` and `scratch mem=0` means the copy-free route is active.
   - `direct_storage=false`, or `direct encode storage allocation was rejected; using scratch+copy`, means the expected portable fallback is active.
5. Stream a motion-heavy scene for 5–10 minutes. There must be no corruption (green/shifted rows, black frames), encoder stalls, or GPU reset.
6. Confirm at least one fallback configuration still streams cleanly. An unsupported capability combination is acceptable and is expected to retain the old scratch-and-copy behavior.

Please record GPU, driver/Mesa version, codec, pixel format, resolution, and whether the direct or fallback route was selected. Vulkan validation-layer warnings should be treated as failures.

JBailes:
#51 is a prerequisite for this, but if we merge both of them, it will break the image build on the second git merge.

Once #51 is merged, we need to revisit this.

PR #53, opened 2026-08-02
fix(comp): subtract the cursor hotspot when rendering client cursor surfaces

dnvang:
## Problem

When a client sets a custom cursor (`wl_pointer.set_cursor`), the compositor renders the cursor surface at the raw pointer location and ignores the hotspot the client submitted with it. Any cursor whose hotspot is not (0,0) is drawn offset by exactly the hotspot, while input lands at the true pointer position.

Conventional arrow cursors put the hotspot at the tip (~0,0), so most apps look fine. Games shipping centre-hotspot cursors are visibly broken: Mindustry (libGDX/arc `Graphics.newCursor`) uses 64x64 cursors with a (32,32) hotspot, so under Wolf its cursor renders 32px down-right of where clicks actually land.

## Fix

Read the hotspot out of the surface's `CursorImageAttributes` and render at `pointer_location - hotspot`, the same way smithay's own anvil example does. (0,0)-hotspot cursors are unaffected. The extra whitespace reflows in the diff are `cargo fmt` reacting to the new lines.

## Verification

Built at current main (b15285a) with the same cargo features as the Wolf image and bind-mounted the resulting `libgstwaylanddisplaysrc.so` over the stock one in the digest-pinned Wolf image: `gst-inspect-1.0 waylanddisplaysrc` reports identically to stock (0.4.0-b15285a), and in Moonlight, Mindustry's cursor now sits exactly on its click target.

### DIFF
--- a/wayland-display-core/src/comp/rendering.rs
+++ b/wayland-display-core/src/comp/rendering.rs
@@ -1,3 +1,4 @@
+use std::sync::Mutex;
 use std::time::{Duration, Instant};
 
 use super::State;
@@ -12,8 +13,9 @@ use smithay::{
         },
     },
     desktop::space::render_output,
-    input::pointer::CursorImageStatus,
+    input::pointer::{CursorImageAttributes, CursorImageStatus},
     render_elements,
+    wayland::compositor::with_states,
 };
 
 pub const CURSOR_DATA_BYTES: &[u8] = include_bytes!("../../resources/cursor.rgba");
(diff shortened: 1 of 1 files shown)


JBailes:
@ABeltramo This is a merge to stable, but it has my approval as is.

ABeltramo:
Good stuff, pretty sure this was reported already somewhere for some specific game but I can't find the original issue (probably a Discord thread then).  
Thanks for fixing!

PR #59, opened 2026-08-19
fix(vulkan): keep the encode-src reuse gate effective across header copies

salty2011:
## Problem

Two holes in the encode-src ring, both in code I added in #45.

`to_gst_buffer()` returns a bare clone of the slot's cached buffer, and copy-on-write defeats the reuse gate. Any element making the header writable at refcount 2 copies it, so the cached slot buffer drops back to refcount 1 while the copy, and the encoder reading through it, still point at the same VkImage. The gate in `convert_inner` then sees the slot as free and lets the producer overwrite an image an encode is still reading.

Two ways in:

- BaseSrc calls `gst_buffer_make_writable()` on the DISCONT frame ("marking pending DISCONT"). For a live source that means start-of-stream and after a flush. Per-frame `do_timestamp` stamping does **not** trigger it, since `gst_base_src_do_sync` writes PTS/DTS directly.
- gwd does not control the downstream graph. An element doing per-buffer `make_writable` (videorate, for one) blinds the gate on **every** frame.

Start-of-stream is not benign either. The first encode is the slowest (warm-up, DPB allocation), and at `RING=1` the next frame reuses the same slot within a frame interval, so the one guaranteed COW lands on the frame most likely still in flight.

Second hole: the gate's 1s timeout logged `"reusing anyway"` and then overwrote the slot. That is the exact hazard the gate exists to prevent, and at `RING=1` there is no other slot to fall back to.

## Change

- `to_gst_buffer` hands out a fresh child header carrying `GstParentBufferMeta` over the slot. The meta's COPY transform re-adds itself and re-refs the parent, so header copies keep the slot non-writable until the last child drops.
- Clearing PTS/DTS/duration on the child gives `do_timestamp` a `NONE` to stamp without touching the shared slot. That replaces the reset `convert_inner` used to do, which is removed.
- The timeout now drops the freshly rendered frame and re-emits the last completed output, rate-limited. The trade is stale-but-correct instead of fresh-but-possibly-corrupt.

## Scope

- `encode_src` only. VA, RGBx and DMABuf keep the bare clone and are byte-identical.
- `parent.copy()` is a shallow header copy: `GstVulkanImageMemory` is not `NO_SHARE` and the range is full, so the memory is ref'd, not duplicated. Zero-copy is intact, and `GST_BUFFER_COPY_META` carries the VideoMeta the encoder needs.

## Verified

`cargo fmt` clean, `cargo check` clean, 23 tests pass.

No runtime reproduction of the hazard. The mechanism is source-level against 1.28 (`gstbasesrc.c`, `gstbuffer.c`, `gstvkimagememory.c`) and the change has production mileage downstream, but I have not demonstrated the gate blinding on hardware. Happy to if you want it before merging.


salty2011:
Ran the reproduction. RADV Renoir, `vulkanh264enc`, RING=4, slow consumer (`identity sleep-time=30000`) so the gate has to engage.

Per-frame gate state, `shared` = slot still referenced at gate entry:

| frame | slot | before | after |
|---|---|---|---|
| 0-3 | 0-3 | `shared=false waited=0` | `shared=false waited=0` |
| **4** | **0** | **`shared=false waited=0`** | **`shared=true waited=82`** |
| 5-9 | 1-3,0,1 | `shared=true waited~200` | `shared=true waited~197` |

Frames 0-3 fill the ring, so no slot is referenced yet. Frame 4 is the first reuse of slot 0, whose frame-0 buffer was the DISCONT frame BaseSrc copy-on-writes. Before the fix that slot alone comes back unreferenced and the gate passes instantly, while every other slot waits ~200 iterations. That is the blinding: one slot, exactly the one that was COW'd. After the fix it waits like the others.

The probe sits inside `if self.encode_src`, so it only emits on the encode-src path at all. Encoder sink pad negotiated `video/x-raw(memory:VulkanImage),NV12,1920x1080`, src pad produced `video/x-h264`, so real frames encoded.

AMD only. I have no NVIDIA host with a Vulkan-encode-enabled build right now, but the mechanism is GstBuffer refcounting and CPU-side gate bookkeeping, so I would not expect vendor variance.

JBailes:
Before writing a ring slot, the producer waits for that slot's GstBuffer to go writable again. Copy-on-write defeats the wait. BaseSrc calls gst_buffer_make_writable() on the DISCONT frame, the header gets copied, the cached slot buffer falls to refcount 1, and the copy still points at the same VkImage. The gate opens on a slot the encoder is reading. GstParentBufferMeta on a child header closes it, because the copy transform re-refs the parent.

So this is the first build where the source genuinely back-pressures, and nobody has measured that. interpipesink never disables enable-last-sample, so it pins the newest raw buffer forever. At RING=4, one slot of four, harmless. At RING=1, the only slot: every frame hits the 1s timeout and the stream runs at 1 fps.

RING=1 is a documented diagnostic. It used to corrupt at full rate. Now it stalls.

Fix README.md:149, then watch busy_drops on hardware. Not for corruption. For hitches.

salty2011:
Fixed the README knob.

`WOLF_VULKAN_RING` now says `1` throttles the stream, names the cause, and gives the workaround:

> The reuse gate waits for a slot's buffer to become writable. A downstream sink holding a reference keeps it non-writable, and `GstBaseSink` leaves `enable-last-sample` on by default, so the sink pins the newest buffer until the next one arrives. At 4 slots the ring rotates past the pinned one. At 1 slot there is nothing to rotate to, so every frame waits the full 1s timeout, drops, and re-emits the previous output. Set `enable-last-sample=false` on the sink when using `1`, or expect about 1 fps.

That is your analysis plus the code path, not a measurement. I have not watched `busy_drops` on hardware yet and did not want to write it up as if I had.

What stopped me: my AMD host reproduces the gate, but I could not get a frame counter working in that image, so I had no fps to report. My NVIDIA host runs NVENC, where the Vulkan encode-src ring is inert. I will come back with RING=1 vs RING=4 from a real session rather than a synthetic pipeline.

Worth flagging separately: we auto-pin RING=1 when HEVC is armed on Vulkan, to dodge the per-slot h265 tiling defect. Same mechanism, so that path likely throttles too. Ours to check, not this PR's.

salty2011:
Correction to my last comment. I did already have `busy_drops` on hardware and failed to connect it.

RING=1 starves with `enable-last-sample` already off. Measured 0.45 fps with busy-drop warnings on an RTX 5090, 2026-07-25. Frame N's child buffer is still downstream (interpipesink to interpipesrc to queue to encoder) when the compositor needs that slot for N+1, so it waits on its own only slot. In-flight refs alone exhaust it. Last-sample retention makes it worse, it is not the whole cause.

That means the workaround I put in the README is wrong. Disabling last-sample does not rescue RING=1. Pushed a correction saying RING=1 is unviable under this gate, full stop.

I pin RING=2 downstream for the same reason. Double-buffering resolves it: h265 59.7 fps, p50 encode 7.0 ms, no green bars, no black-frame cadence.

Separately, I do disable `enable-last-sample` on the source interpipesink, with a unit test asserting it stays off. That was preemptive, not a response to an observed stall.

PR #60, opened 2026-08-19
fix(compositor): close the seat keymap memfd at session teardown

salty2011:
## Problem

smithay mints one sealed `memfd:smithay-keymap` per `Seat::add_keyboard` and keeps it in the `Arc<KbdRc>` behind the seat's `KeyboardHandle`. Dropping `State` closes it only when no grab is left installed.

The cycle:

- The keyboard's own grab slot lives inside that same `KbdRc` (`input/keyboard/mod.rs`, `KbdInternal::grab`).
- Every grab smithay hands out carries a clone of the handle it is installed on. `PopupKeyboardGrab` wraps a `PopupGrab` whose `keyboard_handle` is exactly that clone.
- We install one for every `xdg_popup.grab` (`wayland/handlers/xdg.rs`).

So a grab still active when the session ends is an `Arc` pointing at itself, and the fd stays open for the life of the process. The escape hatches do not save it: the stale-grab discard in `with_grab` and `PopupKeyboardGrab`'s `has_ended()` unset both run on the *next* grab access, which never comes once the session is over.

Measured on a workload whose clients routinely end a session with a popup grab open: 136 back-to-back sessions, exactly one leaked fd each, slope 1.00/cycle with no plateau. Everything else the compositor owns was released on schedule, including the sibling `smithay-dmabuffeedback-format-table` memfd in the same `Display`, which is what rules out a leaked `State` or `Display`. At ~50 sessions a day a host crosses the 1024 fd soft limit in under a month and new sessions fail with `EMFILE`.

## Fix

`State::release_seat`, called once the event loop stops: unset any surviving grab (this is what breaks the cycle), clear keyboard focus, drop the seat's handle. The fd closes deterministically rather than depending on the client having torn its popups down politely.

## Tests

Count the process's open `smithay-keymap` memfds across 8 teardowns, once plainly and once with a grab of `PopupGrab`'s retaining shape installed. Neutering `release_seat` fails the grab test and leaves the plain one passing, which is the expected split since the plain case has no cycle to leak.

The count is process-global while cargo runs this binary's tests in parallel, so the post-cycle sample is the lowest of several readings. A leak is permanent and survives every sample; a neighbouring test's fixture is not. Without that, the two tests contaminate each other's baseline.

## Known gap

Neither test goes through `comp::init`, so the production call site is wired but not covered. Deleting the call would leave the suite green.

## Upstream candidate

The deeper fix belongs in smithay: hold the grab's `KeyboardHandle` weakly, or unset the grab when the last non-grab reference goes away. This is the compositor-side workaround.

## Verified

`cargo fmt` clean, `cargo check` clean, 25 tests pass.

### DIFF
--- a/wayland-display-core/src/comp/mod.rs
+++ b/wayland-display-core/src/comp/mod.rs
@@ -45,7 +45,10 @@ use smithay::{
             backend::{ClientData, ClientId, DisconnectReason, GlobalId},
         },
     },
-    utils::{Clock, DeviceFd, Logical, Monotonic, Physical, Point, Rectangle, Size, Transform},
+    utils::{
+        Clock, DeviceFd, Logical, Monotonic, Physical, Point, Rectangle, SERIAL_COUNTER, Size,
+        Transform,
+    },
     wayland::{
         compositor::{CompositorClientState, CompositorState, with_states},
         dmabuf::{DmabufGlobal, DmabufState},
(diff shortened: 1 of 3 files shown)


JBailes:
@ABeltramo Tag, you are it. Be sure this gets tested.

